PR #1, opened 2016-03-08
Refactor cmake files to use opm infrastructure

jokva:
Borrows macros and common infrastructure from the rest of the opm
project instead of half-implementing its own (as parser must do).
Largely a build system simplification.


joakim-hove:
Tickle travis.

andlaus:
note: instead of closing and immediately re-opening a PR just to force travis to rebuild, you can also log into travis (via github) and press the reload button there...

PR #2, opened 2016-03-16
Moved OutputWriter to opm-output

kjellkongsvik:
Moved opm/core/io/\* in opm-core to opm-output opm/output along with tests.
Is still dependant on opm-core.

This PR is dependant on https://github.com/OPM/opm-common/pull/88


.travis.yml
```diff
@@ -24,10 +24,12 @@ install:
 
     - git clone https://github.com/OPM/opm-common.git
     - git clone https://github.com/OPM/opm-parser.git
+    - git clone https://github.com/OPM/opm-core.git
 
     - opm-parser/travis/clone-and-build-ert.sh
 
     - opm-common/travis/build-opm-common.sh
     - opm-parser/travis/build-opm-parser.sh
+    - opm-parser/travis/build-opm-core.sh
```
bska: This should probably say

```
opm-core/travis/build-opm-core.sh
```

rather than

```
opm-parser/...
```

kjellkongsvik: yes, noticed that :-)

joakim-hove:
OK; I'll merge this now. _Right now_ the situation is:
1. You can download and build the `opm-output` module; but there are no dependencies on the `opm-output` module - i.e. nothing will break&trade;.
2. The code in `opm-output` is a pure copy & paste of the `class OutputWriter` from `opm-core`

The next step will be:
1. A PR removing the `OutputWriter` from `opm-core`.
2. A PR adding a dependency from `opm-autodiff` on `opm-output` 

When the PR's in step two have been merged `opm-output` has become a _hard dependency_.

The current version of `opm-output` will depend on `opm-core`; the medium-term goal is to remove this dependency, but right now `opm-output` needs both the grid and the wells from `opm-core`, so disentangling this dependency is a non-trivial task.

atgeirr:
> right now opm-output needs both the grid and the wells from opm-core

Hoping that opm-grid can serve the need eventually, for wells it may not be so hard -- is it more than well names that are required?

joakim-hove:
> [...] is it more than well names that are required?

For the wells we require:
1. For plotting wells in the restart files we need set of completed cells - that we can have from the `Well` object in the `EclipseState`.
2. To facilitate actual _restart_ we need the `well_state` as seen by the simulator; this might move to e.g. the `SimulationDataContainer` in opm-common`?

PR #6, opened 2016-03-31
Remove unused function argument.

atgeirr:
Silences a warning.

I challenge all developers to use at least -Wall and -Wextra when building OPM!


joakim-hove:
Travis rebuild.

PR #8, opened 2016-04-01
Refactor RFT output

joakim-hove:
With this patch the RFT output writer will write a new RFT file for
every simulation, not update in place which was the case previously.

NB: This requires a very new ert: https://github.com/Ensembles/ert/pull/1036


atgeirr:
Good! Please notify the mailing list when this goes in (due to the ERT requirement change).

PR #10, opened 2016-04-04
copy the `configure` wrapper from opm-autodiff

andlaus:
this enables to build opm-output using an unpatched `dunecontrol` from
Dune 2.3 (and Dune 2.4 if USE_CMAKE is set to "no").

In the medium term the build system needs to become compatible with
Dune's cmake based build system since I think I read somewhere that
the Dune devs intend to remove the autotools based build system before
their next release.

### DIFF
--- a/configure
+++ b/configure
@@ -1,0 +1,35 @@
+#!/bin/sh
+# this file is supposed to be located in the source directory
+src_dir=$(dirname $0)
+
+# scan the arguments and set this if build macros could be specified
+mod_dir=
+for OPT in "$@"; do
+    case "$OPT" in
+        --with-opm-common=*)
+            # remove everything before equal sign and assign the rest
+            mod_dir=${OPT#*=}
+            # tilde expansion; note that doing eval may have side effects
+            mod_dir=$(eval echo $mod_dir)
+            # absolute path
+            [ -d "$mod_dir" ] && mod_dir=$(cd $mod_dir ; pwd)
+            ;;
+    esac
+done
+
+# if it isn't specified, the look around in other known places
+conf_file=cmake/Scripts/configure
+if [ -z "$mod_dir" ]; then
+    if [ -r "$src_dir/$conf_file" ]; then
+        mod_dir="$src_dir"
+    fi
+fi
+
+# terminate with error message here if the module directory is not found
+if [ ! -r "$mod_dir/$conf_file" ]; then
+    echo Build macros not located in \"$mod_dir\", use --with-opm-common= to specify! 1>&2
+    exit 1
+fi
+
+# forward to the corresponding script in the cmake/Scripts/ directory
+exec "$mod_dir/$conf_file" --src-dir="$src_dir" "$@"


bska:
> In the medium term the build system needs to become compatible with Dune's cmake based build system

Why do have to support building Opm using `dunecontrol`?  We do support and should continue to support targeting installed versions of Dune (i.e., treating the Dune core modules as third-party software), but there is no need to adopt the Dune build system for Opm.

andlaus:
> Why do have to support building Opm using dunecontrol?

that's an easy question: because I'm too lazy to spend a full working day banging my head against cmake each time I want to uniformly change the build flags for all Dune and Opm modules for which I need to check something.

(sometimes I do this about 5 times a day.)

> but there is no need to adopt the Dune build system for Opm.

agreed, and this patch does not do this. it just makes it possible to use dune's meta-build tool because currently Opm does not offer even a remotely adequate replacement for `dunecontrol` in my not very humble opinion.

atgeirr:
> it just makes it possible to use dune's meta-build tool

And I'm fine with that and will merge it. It also keeps the modules uniform, since all the others have it. I am not sure what consequences Dune removing autotools support will have though, perhaps someone can start an issue in opm-common for discussion?

alfbr:
> that's an easy question: because I'm too lazy to spend a full working day banging my head against cmake each time I want to uniformly change the build flags for all Dune and Opm modules for which I need to check something.

I have done that daily lately. I have my cmake options in a file, and set the flags with the CMAKE_CXX_COMPILER variable.

akva2:
cmake formalizes this through toolchain files. it's also how you do cross builds and such with cmake, hence it's hidden in http://www.vtk.org/Wiki/CMake_Cross_Compiling

atgeirr:
> I have my cmake options in a file

Same here. Using `cmake -C <options.cmake>` since we started using cmake. I have two option files I use with all modules, one for debug and one for release. Not familiar with toolchain files, is that important?

akva2:
it's the same thing as such. it just allows a bit more, like jailing, which isn't (possible|convenient) through command line options.

andlaus:
Thanks for merging!

> Same here. Using cmake -C <options.cmake> since we started using cmake.

that does not work with the dune modules because they come with a different naming scheme. (note that I don't think that this is better or worse than opm's, just _different_)

the point is that a simple:

``` sh
CMAKE_FLAGS="-DFOO=TRUE"

for MOD in $DUNE_AND_OPM_MODULES; do
  cd $MOD
  cmake $CMAKE_FLAGS
  make
  cd ..
done
```

won't do the trick. (but it does if you use the configure wrapper. Also -- since I have about 10^5 potential combinations of flags/compilers/linkers -- I'm not really eager to create a separate file for each of them.) 

this also does not really deal with the other features that come with dunecontrol, e.g., the automatic dependency resolution or you want to update all modules from git or if you want to run some command in each module's source directory.

blattms:
On Tue, Apr 05, 2016 at 02:16:48AM -0700, Andreas Lauser wrote:

> but that does not work with the dune modules because they come with
> a different naming scheme. (note that I don't think that this is
> better or worse than opm's, just _different_)

Yes, it does. But you have to manually set some of them twice due to
the different naming.

andlaus:
> Yes, it does. But you have to manually set some of them twice due to
> the different naming.

I prefer the configure wrapper in this case.

PR #12, opened 2016-04-05
only update a summary file once per report step

andlaus:
this fixes the problem that well totals are off by a factor of about two. (if there are no drastic changes of the well rates during the substeps.) Note that the better solution would be to continue to write once per substep but skipping report step writes if the adaptive time stepper is used. since I've got no idea how to find this out in opm-output I propose this patch as a stop-gap measure.

this superseeds OPM/opm-core#974, OPM/opm-autodiff#613 and OPM/opm-data#95 . (I suppose my main problem there was that the old output files were not overwritten with these PRs and ctest thus compared the a previously computed result of the master versions.)


joakim-hove:
Ok - this is in flux. Merging now

joakim-hove:
jenkins  build this please

PR #13, opened 2016-04-11
data::Wells format for simulator data exchange

jokva:
Introduces a simple format for data exchange between the simulators and
the the output facilities. Output functions will expect these types, and
it is the simulator's responsibility to create and provide it.

This patch introduces aggregates for well-related information: well
rates, bottom hole pressures and completion rates. Uses a simple bitmask
scheme to differentiate between written values and random garbage.

Supersedes https://github.com/OPM/opm-common/pull/105

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -14,6 +14,7 @@ list( APPEND MAIN_SOURCE_FILES
 
 list (APPEND PUBLIC_HEADER_FILES
         opm/output/OutputWriter.hpp
+        opm/output/Wells.hpp
         opm/output/eclipse/CornerpointChopper.hpp
         opm/output/eclipse/EclipseGridInspector.hpp
         opm/output/eclipse/EclipseIOUtil.hpp
@@ -31,6 +32,7 @@ list (APPEND TEST_SOURCE_FILES
         tests/test_EclipseWriteRFTHandler.cpp
         tests/test_writenumwells.cpp
         tests/test_writeReadRestartFile.cpp
+        tests/test_Wells.cpp
 )
 
 # originally generated with the command:
--- a/opm/output/Wells.hpp
+++ b/opm/output/Wells.hpp
@@ -65,6 +65,7 @@ namespace Opm {
 
         private:
             double& get_ref( opt );
+            const double& get_ref( opt ) const;
 
             opt mask = static_cast< opt >( 0 );
 
@@ -90,21 +91,21 @@ namespace Opm {
 
     /* IMPLEMENTATIONS */
 
-    inline bool Rates::has( opt m ) {
+    inline bool Rates::has( opt m ) const {
         const auto mand = static_cast< enum_size >( this->mask )
                         & static_cast< enum_size >( m );
 
         return static_cast< opt >( mand ) == m;
     }
 
-    inline double Rates::get( opt m ) {
+    inline double Rates::get( opt m ) const {
         if( !this->has( m ) )
             throw std::invalid_argument( "Uninitialized value." );
 
         return this->get_ref( m );
     }
 
-    inline double Rates::get( opt m, double default_value ) {
+    inline double Rates::get( opt m, double default_value ) const {
         if( !this->has( m ) ) return default_value;
 
         return this->get_ref( m );
@@ -130,7 +131,7 @@ namespace Opm {
      * This is an implementation detail and understanding this has no
      * significant impact on correct use of the class.
      */
-    inline double& Rates::get_ref( opt m ) {
+    inline const double& Rates::get_ref( opt m ) const {
         switch( m ) {
             case opt::wat: return this->wat;
             case opt::oil: return this->oil;
@@ -143,6 +144,12 @@ namespace Opm {
                 + std::to_string( static_cast< enum_size >( m ) )
                 + "'" );
 
+    }
+
+    inline double& Rates::get_ref( opt m ) {
+        return const_cast< double& >(
+                static_cast< const Rates* >( this )->get_ref( m )
+               );
     }
 
     }
--- a/opm/output/Wells.hpp
+++ b/opm/output/Wells.hpp
@@ -50,14 +50,14 @@ namespace Opm {
             };
 
             /// Query if a value is set.
-            inline bool has( opt );
+            inline bool has( opt ) const;
 
             /// Read the value indicated by m. Throws an exception if
             /// if the requested value is unset.
-            inline double get( opt m );
+            inline double get( opt m ) const;
             /// Read the value indicated by m. Returns a default value if
             /// the requested value is unset.
-            inline double get( opt m, double default_value );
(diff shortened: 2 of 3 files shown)


CMakeLists_files.cmake
```diff
@@ -14,6 +14,7 @@ list( APPEND MAIN_SOURCE_FILES
 
 list (APPEND PUBLIC_HEADER_FILES
         opm/output/OutputWriter.hpp
+        opm/output/Wells.hpp
```
atgeirr, line 17: Two new files, just one change in the file list. You must add the test program, too.

opm/output/Wells.hpp
```diff
@@ -0,0 +1,173 @@
...
+            enum class opt : enum_size {
+                wat     = (1 << 0),
+                oil     = (1 << 1),
+                gas     = (1 << 2),
+                polymer = (1 << 3),
+            };
+
+            /// Query if a set of values are set. Returns true if all values
+            /// specified by the optmask are set, false if any are missing.
+            inline bool has( opt );
```
atgeirr: To be more clear the method should be named `hasAll`.

opm/output/Wells.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    struct Completion {
+        size_t logical_cartesian_index;
+        Rates rates;
+    };
+
+    struct Well {
+        Rates rates;
+        double bhp;
+        std::map< size_t /*logical_cartesian_index*/, Completion > completions;
```
atgeirr: Cute way to document what the map keys are, but I'd still prefer a normal, explanatory comment...

Edit: or perhaps using a type alias: `using LogicalCartesianIndex = size_t`?

opm/output/Wells.hpp
```diff
@@ -0,0 +1,173 @@
...
+            /// Query if a set of values are set. Returns true if all values
+            /// specified by the optmask are set, false if any are missing.
+            inline bool has( opt );
+
+            /// Read the value indicated by m. Throws an exception if multiple
+            /// values are requested or if the value is unset.
+            inline double get( opt m );
+            /// Read the value indicated by m. Returns errval if multiple
+            /// values are requested or the value is unset.
+            inline double get( opt m, double errval );
```
atgeirr: I think this get() overload should not be public.

CMakeLists_files.cmake
jokva, line 17: You're absolutely right - the test program is in the cmake file on my machine, but I manage to press 'no' when adding it to the commit.

opm/output/Wells.hpp
```diff
@@ -0,0 +1,173 @@
...
+#define OPM_OUTPUT_WELLS_HPP
+
+#include <map>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace Opm {
+
+    namespace data {
```
atgeirr, line 30: Perhaps outputdata would be better? I think the the class names `Rates`, `Well` and `Wells` etc. are nice but very generic, and while they won't collide (since hardly anything else uses the data namespace) it would be better to use a more output-specific namespace. That way we can keep the nice names and be safe that we avoid any confusion.

opm/output/Wells.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    /*
+     * To avoid error-prone and repetitve work when extending rates with new
+     * values, the get+set methods use this helper get_ref to determine what
+     * member to manipulate. To add a new option, just add another case
+     * corresponding to the enum entry in Rates to this function.
+     *
+     * This is an implementation detail and understanding this has no
+     * significant impact on correct use of the class.
+     */
```
atgeirr, line 133: It's true that the user does not have to understand this given sufficient docs etc. But I have to say that I've never seen this pattern before and find it a bit uncomfortable. However, my main focus now is on interface, so I'll ignore this for now...

atgeirr:
In summary, I am mostly fine with this interface. Please also include solvent rate (we need it, so why not put it in straightaway) before it is merged.

I do think the proposed _implementation_ is a bit too clever in places (I do not like the get_ref() and errval things), and I am not sure it is right to let every individual rate has its own mask (perhaps rather on a Report level?). However that can wait (and maybe I'll like it better after some time has passed).

opm/output/Wells.hpp
jokva, line 30: Maybe, even 'output' or 'out' is ok with me. Shorter namespaces are better than longer namespaces though.

opm/output/Wells.hpp
jokva, line 133: It's not really that magical - it's a selector or mapper from some enum to the corresponding variable. It's a static version of `std::map` (or `tuple` without indices), or similar to `namedtuple` in python.

atgeirr:
Please do not rebase during review unless necessary, it tend to break links and make comments unavailable (also I prefer to comment on the diff and not the commits so they do not so easily disappear).

opm/output/Wells.hpp
atgeirr, line 133: I do understand what it does, it's just that one must read it carefully to do so, so it's a bit overkill in terms of cleverness. :-)

akva2:
just in case, @jokva you want to look into git commit --fixup if you aren't aware.

atgeirr:
> into git commit --fixup

Will that solve my reviewer's gripe of losing the messages?

akva2:
yes, github will show them grouped and the original commit is still valid so no comments are lost. it's basically the same as putting a fix commit on top, but with some additional neatness provided by the web interface.

once done, you use git rebase -i &lt;base&gt; --autosquash and git will automagically mark them as fixups.

jokva:
I know it exists, I've just been using --amend since this it's at the top of the history. I can use fixup for single-commit(ish) reviewing purposes, which I presume also works with rebasing for reordering (in case of multi-commit PRs). Any experience there?

akva2:
yeah, that works just fine as long as you don't change the commit messages. it identifies using those to allow reordering. still, you reorder and rebase at the end of the review process, that is the idea.

jokva:
@atgeirr if your comment disappears it means the cause is gone most of the time, doesn't it?

I'll start using fixup.

opm/output/Wells.hpp
jokva, line 133: Maybe. It does however ease maintenance by providing _one_ place that must be updated for supporting a new variable, instead of 3.

atgeirr:
> if your comment disappears it means the cause is gone most of the time, doesn't it?

Not necessarily, often you'd have given a reply before fixing, and it's good to be able to open the outdated diffs to see what was fixed, which is very easy in the github interface until you rebase. I also think that it is better that the discussion is preserved.

jokva:
Rebasing this on master to fix build.

jokva:
A couple of fixups a - few implementation flaws that some testing revealed.

PR #18, opened 2016-04-19
Using getInputGrid API from Parser

pgdr:
Using new EclipseState.getInputGrid API in Parser

See OPM/opm-parser#769


joakim-hove:
jenkins build this opm-parser=769

PR #19, opened 2016-04-22
Well output support

jokva:
Alright, looks like we're approaching a usable state here.

First iteration of user-controlled results output. ERT does most of the heavy lifting, and this patch enables easier use & management from OPM. Comes with a test suite of (most) of the Well-oriented keywords from Norne.

Notable details:
- Intended to be largely invisible to client code, with the notable exception of the timesteps in the simulator (which obviously needs to feed it data for every step).
- It's somewhat `dump-and-forget`, but due to both ERT behaviour and the general behaviour of the SUMMARY format this isn't too bad.
- Unit support (read from EclipseState).

Bugs/needs-for-improvement:
- Doesn't support WPI/WBP/WPB4 as I don't know how to calculate this. @totto82 Could you maybe be of assistance here?
- I'm not certain how we plan to model injector/producer. Currently I assume that if a well is a producer, rates will be positive, and if it is an injector, rates will be negative. The simulator might (probably) views this the other way around, or want to have a dedicated field for injection rates - feedback welcome.

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -1,0 +1,502 @@
+/*
+  Copyright 2016 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <opm/output/eclipse/Summary.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Well.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp>
+#include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+#include <opm/parser/eclipse/Units/ConversionFactors.hpp>
+#include <opm/parser/eclipse/Units/UnitSystem.hpp>
+
+namespace Opm {
+
+namespace {
+
+/*
+ * It is VERY important that the dim enum has the same order as the
+ * metric and field arrays. C++ does not support designated initializers, so
+ * this cannot be done in a declaration-order independent matter.
+ */
+
+enum class dim : int {
+    length,
+    time,
+    density,
+    pressure,
+    temperature_absolute,
+    temperature,
+    viscosity,
+    permeability,
+    liquid_surface_volume,
+    gas_surface_volume,
+    volume,
+    liquid_surface_rate,
+    gas_surface_rate,
+    rate,
+    transmissibility,
+    mass,
+};
+
+namespace conversions {
+    /* lookup tables for SI-to-unit system
+     *
+     * We assume that all values in the report structures are plain SI units,
+     * but output can be configured to use other (inconsistent) unit systems.
+     * These lookup tables are passed to the convert function that translates
+     * between SI and the target unit.
+     */
+
+const double metric[] = {
+    1 / Metric::Length,
+    1 / Metric::Time,
+    1 / Metric::Density,
+    1 / Metric::Pressure,
+    1 / Metric::AbsoluteTemperature,
+    1 / Metric::Temperature,
+    1 / Metric::Viscosity,
+    1 / Metric::Permeability,
+    1 / Metric::LiquidSurfaceVolume,
+    1 / Metric::GasSurfaceVolume,
+    1 / Metric::ReservoirVolume,
+    1 / ( Metric::LiquidSurfaceVolume / Metric::Time ),
+    1 / ( Metric::GasSurfaceVolume / Metric::Time ),
+    1 / ( Metric::ReservoirVolume / Metric::Time ),
+    1 / Metric::Transmissibility,
+    1 / Metric::Mass,
+};
+
+const double field[] = {
+    1 / Field::Length,
+    1 / Field::Time,
+    1 / Field::Density,
+    1 / Field::Pressure,
+    1 / Field::AbsoluteTemperature,
+    1 / Field::Temperature,
+    1 / Field::Viscosity,
+    1 / Field::Permeability,
+    1 / Field::LiquidSurfaceVolume,
+    1 / Field::GasSurfaceVolume,
+    1 / Field::ReservoirVolume,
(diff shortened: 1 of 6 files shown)


joakim-hove:
All in all this is very good; well tested, extendable and (mostly) easy to read. 

Have some minor(?) naming issues which will come in separate comments.

opm/output/eclipse/Summary.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace Opm {
+
+    class EclipseState;
+    class Summary;
+
+namespace out {
+
+class Summary {
+    public:
+        Summary( const EclipseState&, const Opm::Summary& );
```
joakim-hove: Although namespaces technically solve this the naming here too confusing; for a period I was confused by this class which only had copy constructors! Whether this class should get a new name, or the config class from Opm-parser is an open question - my 0.02$ would be to `s/Summary/SummaryConfig/` in opm-parser.

jokva: I'm inclined to agree, considering it essentially only creates a configuration _based_ on the keywords in the summary section, instead of being a representation of the summary section as the name may indicate.

opm/output/eclipse/Summary.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+Summary::Summary( const EclipseState& st,
+                  const Opm::Summary& sum,
+                  const char* file ) :
+    ecl_sum( 
+            ecl_sum_alloc_writer( 
+                file, false, true, ":",
+                to_time_t( st.getSchedule()->getStartTime() ),
+                true,
+                st.getEclipseGrid()->getNX(),
```
joakim-hove: This has been renamed `st.getInputGrid()` upstream.

opm/output/eclipse/Summary.cpp
```diff
@@ -0,0 +1,469 @@
...
+        default: return metric;
+    }
+}
+
+Summary::Summary( const EclipseState& st,
+                  const Opm::Summary& sum,
+                  const char* file ) :
+    ecl_sum( 
+            ecl_sum_alloc_writer( 
+                file, false, true, ":",
```
joakim-hove: The first argument here is not really a _file_ - but a file _stem_, optionally with a leading path prefix. I.e. if passing first argument `"/path/to/CASE"` the summary writer will create files `CASE.SMSPEC` and `CASE.UNSMRY` in `/path/to`. This is typically called _basename_.

joakim-hove: The two `bool` arguments - `fmt_output` and `unified` _should_ be internalized in the `IOConfig` object:

``` C++
const auto& ioConfig = st.getIOConfigConst();
bool fmt_out = ioConfig->getFMTOUT();
bool unified = ioConfig->getUNIFOUT();
...
```

opm/output/eclipse/Summary.cpp
```diff
@@ -0,0 +1,469 @@
...
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Well.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp>
+#include <opm/parser/eclipse/EclipseState/Summary/Summary.hpp>
+#include <opm/parser/eclipse/Units/ConversionFactors.hpp>
+#include <opm/parser/eclipse/Units/UnitSystem.hpp>
+
+namespace Opm {
+
+static const double metric[] = {
```
joakim-hove: Would be good if would infer that these are conversion factors from the code; e.g. 

``` C++
namespace ConversionFactors {
    static const double metric[] = { ... };
    static const double field[] = {...};
}
```

joakim-hove: I guess these should be part of the `anonmyous` namespace - or alternatively the `output` namespace?

opm/output/eclipse/Summary.cpp
```diff
@@ -0,0 +1,469 @@
...
+    gas_surface_volume,
+    volume,
+    liquid_surface_rate,
+    gas_surface_rate,
+    rate,
+    transmissibility,
+    mass,
+};
+
+inline double convert( double si_val, dim d, const double* table ) {
```
joakim-hove, line 115: Can we have `conversion_factors` instead of `table`

opm/output/eclipse/Summary.cpp
```diff
@@ -0,0 +1,469 @@
...
+    const auto x = w.rates.get( phase, 0.0 );
+    switch( phase ) {
+        case rt::gas: return convert( x, dim::gas_surface_volume, dims );
+        default: return convert( x, dim::liquid_surface_volume, dims );
+    }
+}
+
+inline double well_keywords( const smspec_node_type* node,
+                             const ecl_sum_tstep_type* prev,
+                             const double* dims,
```
joakim-hove: I know have given unclear signals regarding the `dims` pointer. The correct&trade; hierarchy is: "Unit system -> Conversion factors" - where the conversion factors are _indexed by dimension_. So we could (I am _not_ suggesting this - just for demonstration) have:

``` C++

class ConversionFactor
public:
  double getConversionFactor( const std::string& systemName , dim d)  {
          const auto& system = conversionFactors.at(system);
          double factor = system.at( d );
          return factor;
  }

private:
   std::map<std::string,std::map<dim , double> > conversionFactors;
}
```

So - would be cool if we could have e.g. `conversion_factors` instead of `dims`; it is now named by the property of the index.

opm/output/eclipse/Summary.hpp
```diff
@@ -0,0 +1,59 @@
...
+        void add_timestep( int report_step, double step_length,
+                           const EclipseState&, const data::Wells& );
+        void write();
+
+    private:
+        ERT::ert_unique_ptr< ecl_sum_type, ecl_sum_free > ecl_sum;
+        std::map< const char*, std::vector< smspec_node_type* > > wvar;
+        const ecl_sum_tstep_type* prev_tstep = nullptr;
+        double duration = 0;
+        const double* dimensions;
```
joakim-hove: `conversion_factors`?

tests/test_Summary.cpp
```diff
@@ -0,0 +1,365 @@
...
+    const auto deck = Parser().parseFile( path, ParseContext() );
+    EclipseState es( deck, ParseContext() );
+    Summary summary( *deck, es );
+    const auto wells = result_wells();
+
+    out::Summary writer( es, summary );
+    writer.add_timestep( 1, 1, es, wells );
+    writer.add_timestep( 2, 1, es, wells );
+    writer.write();
+}
```
joakim-hove, line 358: Would be good to have a test loading the results; i.e. something like:

``` C++
...
writer.write();
{
    ecl_sum_type * sum = ecl_sum_fread_alloc_case( case(?) , ":");
    BOOST_CHECK( ecl_sum_has_key( sum , "WWCT:OPX" );
    ... 
    ecl_sum_free( sum );   
} 

```

jokva, line 358: The loads are already tested (by necessity) via the other, keyword-based tests. Perhaps poorly named, the `WRITE` test is mostly there so that there is a reference output file from the section that can be inspected with other tools, such as the `ert` binaries, for non-automated testing and verification.

joakim-hove, line 358: The testing is _very good and extensive_ on the various variables, actual values and so on - but maybe a bit lacking on the _time direction_? The following `ecl_sum` functions can be used to test/verify some aspects of the time direction:

``` C
/* Complete summary case: */
time_t ecl_sum_get_start_time( const ecl_sum_type * );
time_t ecl_sum_get_data_start( const ecl_sum_type * );
time_t ecl_sum_get_end_time(const ecl_sum_type *);
double ecl_sum_get_sim_length(const ecl_sum_type*); // days


/* Single timestep */
time_t ecl_sum_iget_sim_time( const ecl_sum_type * , int );
double ecl_sum_iget_sim_days( const ecl_sum_type *, int );

/* Report */
time_t ecl_sum_get_report_time( const ecl_sum_type * , int report_step );
```

PR #20, opened 2016-04-26
Group output support

jokva:
A small patchset  that cleans up some clunky behaviour, a bit of refactoring, and support for some of the G\* keywords (those used by Norne).

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -129,7 +129,7 @@ inline std::time_t to_time_t( boost::posix_time::ptime pt ) {
 }
 
 /* The supported Eclipse keywords */
-enum class E {
+enum class E : out::Summary::kwtype {
     WBHP,
     WBHPH,
     WGIR,
@@ -356,7 +356,8 @@ inline double get_vol( const data::Well& w,
     }
 }
 
-inline double well_keywords( const smspec_node_type* node,
+inline double well_keywords( E keyword,
+                             const smspec_node_type* node,
                              const ecl_sum_tstep_type* prev,
                              const double* conversion_table,
                              const data::Well& sim_well,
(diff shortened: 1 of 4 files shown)


opm/output/eclipse/Summary.cpp
```diff
@@ -432,6 +458,93 @@ inline double well_keywords( const smspec_node_type* node,
...
+    switch( phase ) {
+        case rt::wat:
+        case rt::oil: return convert( sum( wells, phase ),
+                                      dim::liquid_surface_volume,
+                                      conversion_table );
+
+        case rt::gas: return convert( sum( wells, phase ),
+                                      dim::gas_surface_volume,
+                                      conversion_table );
+        default: return sum( wells, phase );
```
joakim-hove: Who goes in the `default:`?

jokva: Ideally nothing, and it could be a throw. The fall-back here is just not applying any unit conversions. Preferences?

opm/output/eclipse/Summary.cpp
```diff
@@ -432,6 +458,93 @@ inline double well_keywords( const smspec_node_type* node,
...
+        case E::GOPT: return accu + vol( rt::oil );
+        case E::GGPT: return accu + vol( rt::gas );
+
+        /* Injection rates */
+        case E::GWIR: return - rate( rt::wat );
+        case E::GGIR: return - rate( rt::gas );
+        case E::GGIT: return accu - vol( rt::gas );
+
+        /* Production ratios */
+        case E::GWCT: return rate( rt::wat ) /
```
joakim-hove: Potential division by zero error.

jokva:
Depends on https://github.com/OPM/opm-parser/pull/779

jokva:
The PR has been amended with the following changes:
- Keyword-to-enum mapping is now done once instead of per keyword per timestep. Saves us some computation and opens up for some more interface cleanup later.
- Behaviour in error paths changed - things that **should not happen** unless there is a bug in opm-output throws, and unsupported keyword lookup returns -1.

PR #21, opened 2016-04-28
Remove parameters from eclipse writer

kjellkongsvik:
Dependant on https://github.com/OPM/opm-parser/pull/784


opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -1414,28 +1415,35 @@ void EclipseWriter::writeTimeStep(const SimulatorTimerInterface& timer,
...
     cartesianSize_[1] = eclGrid->getNY();
     cartesianSize_[2] = eclGrid->getNZ();
 
-    if( compressedToCartesianCellIdx ) {
+    numCells_ = eclGrid->getNumActive();
+    gridToEclipseIdx_ = std::vector<int>(numCells_, int(-1) );
+
+    //TODO
+    //This should be calculated in EclipseGrid
+    std::shared_ptr<Opm::GridManager>  ourFineGridManagerPtr = std::make_shared<Opm::GridManager>(eclipseState->getInputGrid());
```
atgeirr: This is quite dangerous. You are building a copy of the grid inside the writer, whereas the outside could possibly use a different mechanism to build the grid (for example flow_mpi.cpp uses the CpGrid class). That could lead to inconsistencies and it can also be a bit expensive. I think that the necessary parts must still be passed in the constructor of EclipseWriter, as before.

joakim-hove: > I think that the necessary parts must still be passed in the constructor of EclipseWriter, as before.

I agree - we must (hopefully temporarily) revert this.

kjellkongsvik:
Good, reverted back those changes. Some cleanup still included. param is not passed. PR https://github.com/OPM/opm-simulators/pull/666 updated for this

bska:
jenkins build this opm-parser=784 please

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -1414,8 +1415,7 @@ void EclipseWriter::writeTimeStep(const SimulatorTimerInterface& timer,
 }
 
 
-EclipseWriter::EclipseWriter(const parameter::ParameterGroup& params,
-                             Opm::EclipseStateConstPtr eclipseState,
+EclipseWriter::EclipseWriter(Opm::EclipseStateConstPtr eclipseState,
                              const Opm::PhaseUsage &phaseUsage,
```
joakim-hove: This way we can get rid of the `phaseUsage` argument:

``` C++
#include <opm/core/props/phaseUsageFromDeck.hpp>
...
Opm::PhaseUsage phaseUsage = phaseUsageFromDeck( eclipseState );
```

akva2:
i think it was this that broke downstreams (opm-simulators), ref https://ci.opm-project.org/job/opm-simulators-new/37/

PR #27, opened 2016-05-09
Added support for writing NNC information

pgdr:
- Writer takes NNC argument
- Sets NNC info to ecl_self_add_nnc
- Writes TRANNNC to INIT
- Writes NNCHEAD, NNC1 and NNC2 to EGRID

Most importantly, `writeInit` takes an optional NNC argument and writes it to file if present.

See
- OPM/opm-simulators#680
- OPM/opm-parser#806

### DIFF
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -42,6 +42,7 @@
 #include <opm/core/props/phaseUsageFromDeck.hpp>
 
 #include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
+#include <opm/parser/eclipse/Units/ConversionFactors.hpp>
 #include <opm/parser/eclipse/Units/Dimension.hpp>
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
 #include <opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp>
@@ -597,11 +598,6 @@ public:
                                         ertPhaseMask(uses),
                                         timer.currentPosixTime());
         }
-
-
-
-
-
     }
 
     void writeKeyword(const std::string& keywordName, const std::vector<double> &data)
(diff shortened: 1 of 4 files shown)


jokva:
Please hold off on this patch for a bit longer as it will introduce a _lot_ of conflicts the refactoring going on.

pgdr:
Let's play chicken.

(Fine, I'll hold…)

jokva:
I had a look and the patch looks great. It's ready to merge, but it's less (total) work if my patches goes through first or this small change is based on my tree rather than current trunk, as essentially all signatures are changed and eclipsewriter itself gets a _substantial_ rework.

If a need for this pops up however and we need it **now** then merge at will.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -1205,6 +1223,13 @@ void EclipseWriter::writeInit(const SimulatorTimerInterface &timer)
             EclipseWriterDetails::restrictAndReorderToActiveCells(data, gridToEclipseIdx_.size(), gridToEclipseIdx_.data());
             fortio.writeKeyword("PERMZ", data);
         }
+        if (nnc.hasNNC()) {
+            std::vector<double> tran;
```
joakim-hove, line 1223: This should be `std::vector<float>`

joakim-hove:
OK - we'll have to resolve the merge order between this and the ongoing `EclipseWriter` reworking, but apart from that I am ready to merge this.

opm/output/eclipse/EclipseWriter.cpp
pgdr, line 1223: Then I need to add an additional `fortio.writeKeyword` that accepts `float` data.  As you can see three lines up, the other data written are also `double`.

joakim-hove, line 1223: > As you can see three lines up, the other data written are also double.

You are right - the `writeKeyword( )` function actually does the `double -> float` conversion.

joakim-hove, line 1223: Need to convert the `tran` vector - which is in SI units, to Eclipse output units. A better unit handling system is in the works from @jokva refactoring, but for now this should do:

``` C++
#include <opm/parser/eclipse/ConversionFactors.hpp>

...

double conversionFactor;
if (eclipseState_->getDeckUnitSystem().getType == UnitSystem::UNIT_TYPE_METRIC)
    conversionFactor = 1.0 / Metric::Transmissibility;
else
   conversionFactor = 1.0 / Field::Transmissibility; 

....

tran.push_back( nd.trans * conversionFactor );
```

jokva, line 1223: It's actually implemented. `conversions::from_si( conversion_table, unit, value )`.

joakim-hove, line 1223: > It's actually implemented. conversions::from_si( conversion_table, unit, value ).

OK - cool!

joakim-hove:
Sorry @jokva

jokva:
Not a problem, my body is ready.

joakim-hove:
> my body is ready.

For a little S & M?

PR #30, opened 2016-05-12
Breaking opm-core dependency in opm-output; new summary implementation

jokva:
Hi,

Let me prefix this patch set which spans several repos by stating the goals:
- We want user-configurable output of summary files etc.
- It would be nice to be able to write eclipse-compatible files outside of the simulator
- We currently have an inverted dependency (output now depends on core, should be the other way around)
- The code itself was rather worn and could use some maintenance. Several features were unused.

With that out of the way, a short synopsis of the major changes:
- The old summary support has been ripped out and replaced with the new (not quite complete) implementation.
- The multiplexing output writer has been removed as it was unused. If there is a need for a similar feature I'll take a fresh approach to this.
- The `OutputWriter` interface has gotten rewritten and does no longer depend on `core` data types.
- A (simple) format for Cell data, called `Solution`, is the information exchange specification required for output operations. Clients are expected to properly populate it.
- I have introduced a temporary utility file in core which assists in the transition. The features of that file are largely naïve and intended as a starting point for internal-to-output data conversion.
- WellState has gotten a virtual method `report` that populates a `data::Wells` structure.
- Internal helpers have been rewritten
- VAG grid support has been removed (completely unused).
- writeVTKdata is **disabled**, other VTK output should work as before.
- EclipseWriter no longer inherits from OutputWriter.

The PR has a lot of code, and changes ripples through both core and simulators; however, I will make sure that with all three HEADs it will compile and tests will run. I'm trying to keep this set a refactoring first of all, so behavioural changes will come later.

I would appreciate if most discussion regarding approach, interfaces etc. happens in this PR, and only direct code comments in the other PRs.

Edit by @joakim-hove:

opm-common: https://github.com/OPM/opm-common/pull/121
opm-core: https://github.com/OPM/opm-core/pull/1016
opm-simulators: https://github.com/OPM/opm-simulators/pull/688


atgeirr:
I have not yet had the time to look at this carefully, and will not be able to until Friday. How urgently do you need feedback?

jokva:
Depends. A lot of things won't work properly yet, but concurrent output is broken in master anyway (and this approach would likely solve some of the fundamental problems).

There's still a lot of work to do and so far they only milestone is breaking up the dependency. Summary output isn't correct (although it should be an easy fix), and several other aspects **will** change.

jokva:
This builds now.



Locally I have modified opm-common to change the dependency order (i.e. core now depends on output), but these changes and test obviously work even if it "depends" on common. Nothing from common is now used in opm-output.

This does **not** mean that all PRs are mergable yet, but now opm-output can be inspected and reviewed more or less on its own.

jokva:
I've changed the PRs in both core and simulators to be à jour with the recent development, and now the simulators and all works. I'd like someone to give travis a hint and build this with OPM/opm-common#121, ~~OPM/opm-parser#814~~, ~~OPM/opm-output#31~~, ~~OPM/opm-output#32~~, OPM/opm-core#1016 and OPM/opm-simulators#688

Oh, and review very welcome.

jokva:
I'd also like to add that the Summary output still isn't complete, but I hope that a switch and some thorough testing  here can uncover the bugs.

opm/output/Cells.hpp
```diff
@@ -0,0 +1,84 @@
...
+            RS,
+            RV,
+        };
+
+        inline bool has( key ) const;
+
+        inline const std::vector< double >& operator[]( key ) const;
+        inline std::vector< double >& operator[]( key );
+
+        void insert( key, std::vector< double > );
```
joakim-hove, line 51: Shouldn't this be: `void insert( key, std::vector< double > &&);`

opm/output/Cells.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+        /* data::Solution expect the following to assumptions to be true:
+         * * vector index corresponds to cell index
+         * * all units are SI
+         * * cells are active indexed
+         */
+        std::map< key, std::vector< double > > data;
```
joakim-hove, line 58: I guess this should be private? At least when `has()` and `operator[]` has been implemented.

opm/output/Cells.hpp
joakim-hove, line 51: For sanity checks it would make sense to check the size of the input vector; the expected number of cells could be a member variable specified in the constructor.

opm/output/Cells.hpp
jokva, line 58: Maybe, I'm not sure yet. I initially just aliased `std::map` to Wells, but later needed more so I wrapped a class around it. Now it should possibly be private, but it's exposed in order to be data "dumping" format (so to speak). I figured the answer would be clearer once everything started calming down around it, and this is good feedback w.r.t. that.

opm/output/Cells.hpp
jokva, line 51: Doesn't have to be, and doesn't mandate it. The operation itself will nearly always be a move anyway (especially if I get my shit together and make the definition inline which it should've been in the first place).

edit: but half the time this is the same vector that the simulator keeps around, so that just moves the copy burden over on them (which isn't great). An overload if anything.

opm/output/OutputWriter.hpp
```diff
@@ -74,44 +75,28 @@ class OutputWriter {
      * This routine should be called before the first timestep (i.e. when
      * timer.currentStepNum () == 0)
      */
-    virtual void writeInit(const SimulatorTimerInterface &timer, const NNC& nnc = NNC()) = 0;
+    virtual void writeInit( const NNC& nnc = NNC() ) = 0;
```
joakim-hove: Has been much fighting with this NNC - please avoid default variable for now.

opm/output/OutputWriter.hpp
```diff
@@ -52,10 +53,10 @@ struct PhaseUsage;
  *          OutputWriter::create (params, parser);
  *
  *  // before the first timestep
- *  writer->writeInit (timer);
+ *  writer->writeInit( current_posix_time, time_since_epoch_at_start );
```
joakim-hove, line 56: `time_since_epoch_at_start` ~ is `seconds_elapsed` a better name.

jokva, line 56: Old patch. The entire thing is gone.

opm/output/OutputWriter.hpp
jokva: Inherited and not changed, but I can remove it for sure.

opm/output/OutputWriter.hpp
```diff
@@ -74,44 +75,28 @@ class OutputWriter {
...
-    virtual void writeTimeStep(const SimulatorTimerInterface& timer,
-                               const SimulationDataContainer& reservoirState,
-                               const WellState& wellState,
-                               bool  isSubstep) = 0;
+    virtual void writeTimeStep( int report_step,
+                                time_t current_posix_time,
+                                double seconds_elapsed,
+                                data::Solution reservoirState,
+                                data::Wells,
+                                bool  isSubstep) = 0;
```
joakim-hove, line 98: We are semantically not really interested if this is a substep or not; the interesing question is:
1. Should I write summary data?
2. Should I write restart data?

The answer to the first question is "always" yes; although you can use the keyword `RPTONLY`[1] to essentially configure a `false` on this. For the second question we have logic like:

``` C++
if (!isSubstep)
   printRestart();
```

But then I think it is better to name the variable something like `printRestart` - which is the real question; in exceptional cases that can also be set to true even for a substep - although I do not think our full stack supports that at the moment.

[1]. I do not think we support that - quite uncommon.

opm/output/OutputWriter.hpp
pgdr: Is that necessary?  The fighting with NNC has not been caused by this.  I furthermore think that a default NNC here makes sense, since if we don't have NNC information to write to output, it shouldn't be up to the caller of this method to create an empty NNC object.

opm/output/OutputWriter.hpp
jokva, line 98: I totally agree and it's a much better interface.

I'll make a note of it and add it to the list of interface improvements (the `isSubstep` logic is inherited from the old implementation).

opm/output/Wells.hpp
```diff
@@ -150,7 +176,43 @@ namespace Opm {
     inline double& Rates::get_ref( opt m ) {
         return const_cast< double& >(
                 static_cast< const Rates* >( this )->get_ref( m )
-               );
+                );
+    }
+
+    inline Well& Wells::operator[]( const std::string& k ) {
```
joakim-hove, line 182: Do we want the autovivification properties of `operator[]`?

opm/output/OutputWriter.hpp
jokva: Quick check in simulators shows that the NNC is always passed regardless, although the discussion probably should go on.

opm/output/Wells.hpp
jokva, line 182: Maybe. It exists for clients so they have an easier time building in the case of many wells. I have been playing with the idea of forbidding to modify an established `Wells` though, i.e. all wells (and by extension connections) must be passed at construction time.

opm/output/Wells.hpp
```diff
@@ -88,7 +89,32 @@ namespace Opm {
...
+        inline Wells( std::initializer_list< value_type >,
+                      std::vector< double > bhp,
+                      std::vector< double > temperature,
+                      std::vector< double > wellrates,
+                      std::vector< double > perf_pressure,
+                      std::vector< double > perf_rates );
+
+        std::map< std::string, Well > wells;
+        std::vector< double > bhp;
+        std::vector< double > temperature;
```
joakim-hove, line 113: The temperature variable is orders of magnitude less relevant than the others. In most(?) of the cases it will only have a fixed dummy value - if not downright removed it should at least be removed to the end of the argument list.

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -17,230 +17,163 @@
...
-            if (ecl_file_has_kw(file_type, sgas)) {
-                ecl_kw_type* sgas_kw = ecl_file_iget_named_kw(file_type , sgas, 0);
-                sgas_data = ecl_kw_get_float_ptr(sgas_kw);
-                std::vector<double> sgas_datavec(&sgas_data[0], &sgas_data[numcells]);
-                EclipseIOUtil::addToStripedData(sgas_datavec, simulator_state.saturation(), phaseUsage.phase_pos[BlackoilPhases::Vapour], phaseUsage.num_phases);
-            } else {
-                std::string error_str = "Restart file is missing SGAS data!\n";
-                throw std::runtime_error(error_str);
-            }
+        struct keyword {
```
joakim-hove: Hopefully you could benefit from using: https://github.com/Ensembles/ert/blob/master/devel/libeclxx/include/ert/ecl/EclKW.hpp

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -17,230 +17,163 @@
...
-            if (ecl_file_has_kw(file_type, swat)) {
-                ecl_kw_type* swat_kw = ecl_file_iget_named_kw(file_type , swat, 0);
-                swat_data = ecl_kw_get_float_ptr(swat_kw);
-                std::vector<double> swat_datavec(&swat_data[0], &swat_data[numcells]);
-                EclipseIOUtil::addToStripedData(swat_datavec, simulator_state.saturation(), phaseUsage.phase_pos[BlackoilPhases::Aqua], phaseUsage.num_phases);
-            } else {
-                std::string error_str = "Restart file is missing SWAT data!\n";
-                throw std::runtime_error(error_str);
-            }
+        for( const auto* key : { "PRESSURE", "TEMP", "SWAT", "SGAS" } ) {
```
joakim-hove, line 43: This must be configurable; see previous comment about "TEMP". For the phases present the `phaseUsage.phase_used[BlackoilPhases::Aqua]` or similar should be consulted.

General comment:
1. The model always(?) has a water phase - i.e. `SWAT`.
2. A two-phase model will _only_ have water - the saturation of the _other_ phase, be that oil or gas, is implicitly inferred from `S_{other} = 1 - S_{wat}`.
3. For three phase models we enumarate water and gas saturations; if needed the oil saturation is `S_{oil} = 1 - S_{gas} - S_{wat}`.

opm/output/Wells.hpp
jokva, line 113: Maybe. This is slightly a preservation of the old approach, but more importantly the `bhp`, `temperature` and `wellrates` must all have the same size, where `perf` don't.

I do agree the constructor needs some work though.

opm/output/eclipse/EclipseReader.cpp
jokva, line 43: `PhaseUsage` is unavailable, and this is currently just a port of old behaviour. I've added it on the TODO list.

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -17,230 +17,163 @@
...
+        const auto apply_temperature = [=]( double x ) {
+            return units.to_si( UnitSystem::measure::temperature, x );
+        };
+
+        std::transform( sol[ ds::PRESSURE ].begin(), sol[ ds::PRESSURE ].end(),
+                        sol[ ds::PRESSURE ].begin(), apply_pressure );
+        std::transform( sol[ ds::TEMP ].begin(), sol[ ds::TEMP ].end(),
+                        sol[ ds::TEMP ].begin(), apply_temperature );
+
+        /* optional keywords */
```
joakim-hove, line 85: The `RS`and `RV`keywords are not really optional - if the model is specified with the `DISGAS` and `VAPOIL` keywords these keywords _should_ be present in the restart file.

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -17,230 +17,163 @@
...
+    inline data::Wells restoreOPM_XWEL( ecl_file_type* file,
+                                        int num_wells,
+                                        int num_phases ) {
+        const char* keyword = "OPM_XWEL";
+
+        ecl_kw_type* xwel = ecl_file_iget_named_kw( file, keyword, 0 );
+        const double* xwel_data = ecl_kw_get_double_ptr(xwel);
+        const double* xwel_end  = xwel_data + ecl_kw_get_size( xwel );
+
+        const double* bhp_begin = xwel_data;
```
joakim-hove, line 108: Holy diver - I thought pointer arithmetic was frowned upon?

jokva, line 108: Is it? They're iterators in this context.

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -17,230 +17,163 @@
...
+
+std::pair< data::Solution, data::Wells >
+init_from_restart_file( const EclipseState& es, int numcells ) {
+
+    InitConfigConstPtr initConfig        = es.getInitConfig();
+    IOConfigConstPtr ioConfig            = es.getIOConfig();
+    int restart_step                     = initConfig->getRestartStep();
+    const std::string& restart_file_root = initConfig->getRestartRootName();
+    bool output                          = false;
+    const std::string& filename          = ioConfig->getRestartFileName(
```
joakim-hove: The `getRestartFileName()` has value return (should probably a have a better name ....); so I think the filename should be a value as well.

jokva: It does! I'll update the patch.

opm/output/Wells.hpp
joakim-hove, line 113: > but more importantly the bhp, temperature and wellrates must all have the same size

I mess up this every second time; but I would have thought: `wellrates.size() == bhp.size() * numphases`?

opm/output/eclipse/EclipseReader.cpp
joakim-hove, line 108: > They're iterators in this context

They are used as iterators - but surely pointer arithmetic is used; as long as it is tested it is fine by my though.

opm/output/eclipse/EclipseReader.cpp
joakim-hove, line 43: > PhaseUsage is unavailable, and this is currently just a port of old behaviour. I've added it on the TODO list.

OK; the essential information in the `phaseUsage` variable is available in `EclipseState` as well.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -104,171 +77,98 @@ void restrictAndReorderToActiveCells(std::vector<double> &data,
...
-        for (int i = 0; i < numEntries; ++i) {
-            target[i] = static_cast<T>(data[i]);
-        }
+        for( size_t i = 0; i < data.size(); ++i )
+            target[ i ] = T( data[ i ] );
     }
 
-    void set(const std::string name, const std::vector<const char *>& data)
+    Keyword( const std::string& name, const std::vector< const char* >& data ) :
+        handle( ecl_kw_alloc( name.c_str(), data.size(), ECL_CHAR_TYPE ) )
```
joakim-hove: This `handle` variable

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -104,171 +77,98 @@ void restrictAndReorderToActiveCells(std::vector<double> &data,
...
-        if (std::is_same<T, const char *>::value)
-        { return ECL_CHAR_TYPE; }
-
-
-        OPM_THROW(std::logic_error,
-                  "Unhandled type for data elements in EclipseWriterDetails::Keyword");
-    }
-
-    ecl_kw_type *ertHandle_;
+    ERT::ert_unique_ptr< ecl_kw_type, ecl_kw_free > handle;
```
joakim-hove: Again - here I hope you can use: https://github.com/Ensembles/ert/blob/master/devel/libeclxx/src/EclKW.cpp - if not we should improve the ert class.

If it does not make sense to use the ert class please find a better name than `handle`[1] - handle to what??

[1]: I know the old used that name - and it always annoyed me.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -104,171 +77,98 @@ void restrictAndReorderToActiveCells(std::vector<double> &data,
...
-                                             baseName.c_str(),
-                                             type,
-                                             formatted,
-                                             writeStepIdx);
-    }
-
-    ~FileName()
-    { std::free(ertHandle_); }
+             bool formatted ) :
+        handle( ecl_util_alloc_filename(
```
joakim-hove: And now `handle` is suddenly a `char *`\- what a mess this was.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -296,277 +196,151 @@ class Restart : private boost::noncopyable
...
-    ecl_rst_file_type *ertHandle() const
-    { return restartFileHandle_; }
+    ecl_rst_file_type* ertHandle() { return this->handle.get(); }
+    const ecl_rst_file_type* ertHandle() const { return this->handle.get(); }
 
 private:
-    char *restartFileName_;
-    ecl_rst_file_type *restartFileHandle_;
+    FileName filename;
+    ERT::ert_unique_ptr< ecl_rst_file_type, ecl_rst_file_close > handle;
```
joakim-hove: Again: `handle`

opm/output/eclipse/EclipseReader.cpp
jokva, line 108: In C++, pointers are a special case of iterators, implemented as a `ptrsize` integer (or whatever that constant is called, I never remember).

I've made a note of it though, and I'll consider changing it into something less surprising.

opm/output/eclipse/EclipseWriter.cpp
jokva: That's a nice addition, I'll look into it now.

opm/output/eclipse/EclipseReader.cpp
jokva, line 43: That's true, and it's where I get the phase information from now.

opm/output/Wells.hpp
jokva, line 113: That's true, an oversight on my part. Which is why I put it after the `temperature` so the vectors are grouped by expected size.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -104,171 +77,98 @@ void restrictAndReorderToActiveCells(std::vector<double> &data,
...
+                std::free )
+    {}
 
-    const char *ertHandle() const
-    { return ertHandle_; }
+    const char* ertHandle() const { return this->handle.get(); }
 
 private:
-    char *ertHandle_;
+    using fd = std::unique_ptr< char, decltype( std::free )* >;
```
joakim-hove, line 123: Better naming required - `fd == filedescriptor`; and again `handle`??

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -296,277 +196,151 @@ class Restart : private boost::noncopyable
...
-                                                   type_of_restart_file,
-                                                   ioConfig->getFMTOUT(), // use formatted instead of binary output?
-                                                   writeStepIdx);
+            const IOConfig& ioConfig ) :
+        filename( outputDir,
+                baseName,
+                ioConfig.getUNIFOUT() ? ECL_UNIFIED_RESTART_FILE : ECL_RESTART_FILE,
+                writeStepIdx,
+                ioConfig.getFMTOUT() ),
+        handle(
```
joakim-hove: Here `handle`seems to be a `ecl_file_type *` instance?

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -296,277 +196,151 @@ class Restart : private boost::noncopyable
...
+    Init( const std::string& outputDir,
+          const std::string& baseName,
+          int writeStepIdx,
+          const IOConfig& ioConfig ) :
+        egrid( outputDir,
+               baseName,
+               ECL_EGRID_FILE,
+               writeStepIdx,
+               ioConfig.getFMTOUT() ),
+        init( fortio_open_writer(
```
joakim-hove: Here you can look at: https://github.com/Ensembles/ert/blob/master/devel/libeclxx/include/ert/ecl/FortIO.hpp#L35

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -296,277 +196,151 @@ class Restart : private boost::noncopyable
...
+    void writeHeader( int numCells,
+                      const int* compressedToCartesianCellIdx,
+                      time_t current_posix_time,
+                      const EclipseState& es,
+                      int ert_phase_mask,
+                      const NNC& nnc = NNC() )
     {
-        auto dataField = eclipseState->get3DProperties().getDoubleGridProperty("PORO").getData();
-        restrictAndReorderToActiveCells(dataField, numCells, compressedToCartesianCellIdx);
+        auto dataField = es.get3DProperties().getDoubleGridProperty("PORO").getData();
```
joakim-hove, line 283: We really should leave it as is for now - to get this merged; but quite soon I would like to changes this:

``` C++
const auto& dataField = es.get3DProperties().getDoubleGridProperty("PORO").getData();
auto outputField = restrictAndReorderToActiveCells( dataField, numCells, compressedToCartesianCellIdx );
```

The reseating taking place in `restrictAndReorderToActiveCells()` is less than intuitive.

opm/output/eclipse/EclipseReader.cpp
jokva: Yes, but it requires patching of EclKW. It's coming.

opm/output/eclipse/EclipseWriter.cpp
jokva, line 123: `file descriptor` as is common in linux programs. Anywho, it was mostly to make the variable declaration neater, and no biggie.

opm/output/eclipse/EclipseWriter.cpp
jokva, line 283: I think the goal should be to get rid of restrict-and-reorder altogether, but that's on the horizon, and as you rightly say, out of the scope of this change set.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -580,880 +354,418 @@ class Init : private boost::noncopyable
...
 
-    fortio_type *ertHandle() const
-    { return ertHandle_; }
+    fortio_type* ertHandle() { return this->init.get(); }
 
 private:
-    fortio_type *ertHandle_;
-    FileName egridFileName_;
+    FileName egrid;
+    ERT::ert_unique_ptr< fortio_type, fortio_fclose > init;
```
joakim-hove: The name `init` says nothing about this being a file-like object. I would maybe assume it did not even need to be member -  but that can be handled in the future; the same applies to the `FileName egrid`.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -580,880 +354,418 @@ class Init : private boost::noncopyable
...
 };
 
+const int inactive_index = -1;
 
+/// Convert OPM phase usage to ERT bitmask
```
joakim-hove, line 345: The `ECL_WATER_PHASE` and so on values are really Eclipse values and not ERT values.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -580,880 +354,418 @@ class Init : private boost::noncopyable
...
-        // average for the time step. For more complicated time stepping schemes, the
-        // integral of the rate is not simply multiplying two numbers...
-        const double intg = timer.stepLengthTaken() * rate(wellState);
-
-        // add this timesteps production to the total
-        total_ += intg;
-        // report the new production total
-        return unit::convert::to(total_, targetRateToSiConversionFactor_);
-    }
+    for( size_t active_index = 0; active_index < num_cells; ++active_index )
```
joakim-hove, line 396: We are maintaning varieties of this mapping again and again; for the future we should try to use _only_ the grid - all the required mappings are there.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -580,880 +354,418 @@ class Init : private boost::noncopyable
...
-        }
-        else if (unitType == UnitSystem::UNIT_TYPE_METRIC) {
-            unitName_ = "BARSA";
-            targetRateToSiConversionFactor_ = barsa; // Pa -> bar
-        }
-        else
-            OPM_THROW(std::logic_error,
-                      "Unexpected unit type " << unitType);
+            const auto global_index = grid.getGlobalIndex( i, j, k );
+            const int index = this->global_to_active[ global_index ];
```
joakim-hove:
``` C++
if (!grid.cellActive(i,j,k))
   continue;
```

joakim-hove: If https://github.com/OPM/opm-parser/pull/817 is merged you could do:

``` C++
if (!grid.cellActive(i,j,k))
   continue;
{
     size_t index = grid.activeIndex(i,j,k);
     const double depth = ...
     const double press = !pressure.empty() ? pressure[ index ] : 0.0;
     .....
}
```

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -580,880 +354,418 @@ class Init : private boost::noncopyable
...
-          break;
-        default:
-          ert_well_type = IWEL_UNDOCUMENTED_ZERO;
-      }
-  }
-
-  return ert_well_type;
 }
 
+class EclipseWriter::Impl {
```
joakim-hove, line 463: Please - can we stop this pattern?!

opm/output/eclipse/EclipseWriter.cpp
andlaus, line 283: note that outputing _anything_ directly from EclipseState's grid properties should be considered to be bad style because these properties may be inconsistent with what the simulator actually uses: in this case (PORO), the simulator actually uses the pore volume which may or may not be calculated using PORO; in any case, the porosity gets modified by the rock compressibility, NTG and MULTPV, so the values which are put into the output files might not be what the user expects...

opm/output/eclipse/EclipseWriter.cpp
jokva, line 463: What, the private implementation? It saves us exposing all the internal RFT/Restart/Init etc. to users, which is nice.

If you're talking about the name itself I don't have that strong an opinion (since it's isolated anyway), but I figured it was better than my original idea of "E" or "EW".

opm/output/eclipse/EclipseWriter.cpp
jokva, line 283: This is for the INIT file though in which case we're interested in writing whatever is in the Deck (no?), rather than what the simulator sets up internally later.

Correct me if I'm wrong in this, but I'd assume that meant outputting whatever is provided by the Deck (and so EclipseState).

joakim-hove, line 283: There is certainly room for improvement in this code; the current ambition is "just" to get it to compile and work with the opm-core <-> opm-output depency changed (i.e. opm-output should _not_ depend on opm-core). When 
that is fixed we can continue with improvements, but your initial claim:

> note that outputing anything directly from EclipseState's grid properties should be considered to be bad style because 

Is just plain wrong. 

You are right that the simulator uses `PORV` - which is either initialized from cell volumes, `PORO`and `NTG` or entered explicitly in the grid. We should output `PORV` - from the `Eclipse3DProperties` array[1]; and if present in the grid we should output `PORO` and also `NTG`. But the fact the simulator uses `PORV` does not in any way warrant the claim that we should not output `PORO`.

[1]: Actually the `PORV` vector is somewhat special in the INIT file, that vector _should_ be written as a vector of `nx * ny * nz` elements - where zero values can be used to infer the active-inactive cells without using a grid. We do not have that special casing in place yet (I think) - but that should be trivial to add.

andlaus, line 283: I thought that the intention of the ECL restart files is to output what the simulator uses and not repeat what the deck says. Given that this is ECL, I'm not too much surprised if that is not the case, though. (I'm just wondering what the value proposition of that data is then.)

> You are right that the simulator uses PORV - which is either initialized from cell volumes, POROand NTG or entered explicitly in the grid. We should output PORV - from the Eclipse3DProperties array[1]; and if present in the grid we should output PORO and also NTG. But the fact the simulator uses PORV does not in any way warrant the claim that we should not output PORO.

do you mean that an additional check for PORO is needed because it is not present in some decks? In that case, I agree, but what happens for "partially defaulted" PORO fields where the defaulted cells are specified via PORV?

tests/test_Restart.cpp
```diff
@@ -0,0 +1,300 @@
...
+            snd.second.perf_pressure.begin(), snd.second.perf_pressure.end()
+        );
+    BOOST_CHECK_EQUAL_COLLECTIONS(
+            fst.second.perf_rate.begin(), fst.second.perf_rate.end(),
+            snd.second.perf_rate.begin(), snd.second.perf_rate.end()
+        );
+}
+
+BOOST_AUTO_TEST_CASE(EclipseReadWriteWellStateData) {
+    test_work_area_type * test_area = test_work_area_alloc("test_Restart");
```
joakim-hove, line 293: This is available as: https://github.com/Ensembles/ert/blob/master/devel/libert_utilxx/include/ert/util/TestArea.hpp

jokva:
norne-snap-output

edit: looks like there might be an off-by-one error in rates here, I'll look into this.

jokva:
This should be ready to merge now.

norne-snap-output1

bska:
This PR breaks the build of module opm-upscaling.  Given the magnitude of the change that is probably not entirely unexpected but we still need to fix it.  In particular, class [`Opm::SteadyStateUpscalerImplicit<>`](https://github.com/OPM/opm-upscaling/blob/master/opm/upscaling/SteadyStateUpscalerImplicit.hpp#L52) is not properly converted to account for changes in the [`writeECLData`](https://github.com/OPM/opm-output/blob/master/opm/output/eclipse/writeECLData.hpp#L32-L39) interface of commit 43ce7fc.

I do not know if there are other build issues in module opm-upscaling.

jokva:
I'll have a look, but it hasn't changed behaviour so it should just be an interface upgrade.

jokva:
@bska https://github.com/OPM/opm-upscaling/pull/194 Mind checking it out?

PR #32, opened 2016-05-23
Write rates/totals correctly

jokva:
Now assumes clients hands over production rates as negative injection (i.e. all rates are injection rates, and production rates are determined by rates being negative). Summary should now write correct values for what it actually writes.

Along with that, injection rate tests are added. Finally, I've grouped tests on whether or not they work on well- or group keywords rather than on immediate family (production rate, total, injection etc).

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -381,16 +381,16 @@ inline double well_keywords( E keyword,
     switch( keyword ) {
 
         /* Production rates */
-        case E::WWPR: return rate( rt::wat );
-        case E::WOPR: return rate( rt::oil );
-        case E::WGPR: return rate( rt::gas );
-        case E::WLPR: return rate( rt::wat ) + rate( rt::oil );
+        case E::WWPR: return - rate( rt::wat );
+        case E::WOPR: return - rate( rt::oil );
+        case E::WGPR: return - rate( rt::gas );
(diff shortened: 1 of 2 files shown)


opm/output/eclipse/Summary.cpp
```diff
@@ -548,21 +546,21 @@ inline double group_keywords( E keyword,
...
+        case E::GGPR: return - rate( rt::gas );
+        case E::GLPR: return (- rate( rt::wat )) + (- rate( rt::oil ));
 
         /* Production totals */
-        case E::GWPT: return accu + vol( rt::wat );
-        case E::GOPT: return accu + vol( rt::oil );
-        case E::GGPT: return accu + vol( rt::gas );
+        case E::GWPT: return accu + ( - vol( rt::wat ) );
+        case E::GOPT: return accu + ( - vol( rt::oil ) );
+        case E::GGPT: return accu + ( - vol( rt::gas ) );
```
joakim-hove: Simplify the math:

``` C++
case E::GLPR: return - (rate( rt::wat )) + rate( rt::oil )); 
...
case E::GGPT: return accu - vol( rt::gas ) ;
```

jokva: Sure. I did it this way to emphasise that accumulation is increasing and the result is positive, but I don't mind doing it the other way (or maybe later do an abs()).

PR #35, opened 2016-05-25
Use class scope (::) rather than access enum as member.

atgeirr:
This restores compilation with clang for me.

I am not sure about the legality of the original syntax, but find the new one a bit clearer in any case (no risk of confusing with a data member).


jokva:
Standard §5.2.5 (4.5) 

> If E2 is a member enumerator and the type of E2 is T, the expression E1.E2 is a prvalue. The type of E1.E2 is T.

PR #39, opened 2016-06-13
Minor simplifications and removal of ReorderAndRestrictToActiveCells

jokva:
A handful of deletions and transformations towards better output support. Key changes
- OutputWriter.hpp is deleted as it is completely unused. It was scheduled for removal for quite some time now and here it is.
- output only uses boost for unit tests.
- `convertFromSiTo` has been replaced in favour of `std::transform`
- `ReorderAndRestrictToActive` is scheduled for complete removal, and the plan is to use `EclipseGrid` to manage the relevant information. Regardless, most use of this function is unnecessary and confusing, and more often than not it's a no-op. This requires a few small changes in opm-simulators and some testing of MPI runs with (with restart).

Thus, I consider running MPI-Norne a part of this review, and I'd also like someone to take a look at the generated INIT files both in sequential runs and in parallel.

All tests pass both here and in simulator on my machine.


bska:
> `ReorderAndRestrictToActive` is scheduled for complete removal, and the plan is to use `EclipseGrid` to manage the relevant information.

That's just not possible.

jokva:
Why not? With this patch, the only use of restrictAndReorder is the removal of inactive cells from the grid for PERM(XYZ) and to set up the active/inactive cells via ACTNUM in the INIT file header. Can't this information be passed by modifying a grid instance directly and pass the modified grid instance to the writeInit function?

bska:
> to set up the active/inactive cells via ACTNUM in the INIT file header.

And that's the bit that generally isn't correct.  Unless `EclipseGrid` learns how to process `COORD` and `ZCORN` in **exactly** the same way as the simulator (including pinch, net-to-gross, `MINPV`), `ACTNUM` is not sufficient to determine the set of active cells.

jokva:
The idea is that the simulator should set these details accordingly, and then pass that (now modified) grid to output.

bska:
> The idea is that the simulator should set these details accordingly, and then pass that (now modified) grid to output.

So the `EclipseGrid` is scheduled to become mutable?

jokva:
No. The simulator would, in case its structure differs from the one provided by `EclipseState`, copy the grid instance, perform **local mutation** on its local copy, then pass that copy to `writeInit`. If no such modification is done then you can pass the `EclipseState` provided instance.

We'll eventually get to the same thing with `IOConfig` and other overriden-deck-provided-properties.

bska:
> The simulator would, in case its structure differs from the one provided by `EclipseState`

The simulator must always assume that that is the case, so every simulator run must start by duplicating the `EclipseGrid`.

jokva:
Probably yes.

jokva:
I'd like to point out that the `RestrictAndReorder` haven't really done anything in sequential runs for quite some time, and unless I miss something cruicial even the MPI runs gather to the master node before calling a sequential write, so the `RestrictAndReorder` appears to always be a no-op.

bska:
> I'd like to point out that the `RestrictAndReorder` hasn't really done anything in sequential runs for quite some time,

Okay, then I'm obviously missing something because I see no other place that does `global->active` subset extraction for output purposes.

jokva:
Sure, but global->active mapping was only performed when the passed `compressedToCartesianCellIdx` array was non-nullptr which it only was in parallel runs. This turns the function into a no-op (or mapping indices on their identity).

I might've misunderstood something, but I haven't been able to break anything. _Yet_.

bska:
> Sure, but global->active mapping was only performed when the passed `compressedToCartesianCellIdx` array was non-nullptr which it only was in parallel runs. This turns the function into a no-op (or mapping indices on their identity).

That's a bug in the simulator, then.  The global cell mapping is hardly ever null, particularly not in a corner-point model.

jokva:
You seem to be right - `ParallelDebugOutput` seems to be used regardless of the run being parallel (and debug) and provides the grid mapping. That grid mapping is still passed to the Init writing function, so it should work as expected.

The real difference is in restart/summary writing, in which case the revised assumption is that the Simulator provides only active and cartesian indexed cells.

joakim-hove:
> So the EclipseGrid is scheduled to become mutable?

The `EclipseGrid`is already - to some extent - mutable in the sense that the set of active cells can be modified with `EclipseGrid::resetACTNUM()` - that is invoked before the grid is written to disk. The current grid initialization is as follows:
1. The grid keywords are loaded from disk and an `EclipseGrid` instance is created as part of the `EclipseState`.
2. The `EclipseGrid` is passed to simulator grid structures, and used as foundation for either a `struct UnstructuredGrid`or `Dune CornerPoint`. The grid preprocessors due their `PINCH, MINPV+++`magic and determine an updated `ACTNUM`value.
3. Simulation runs ....
4. When writing en `EGRID` file the `EclipseGrid` instance is modified with updated `ACTNUM` and then written to disk.

In the pipeline is changes to be able to instantiate an `EclipseGrid` from the simulation grid, so that we ensure that the output grid has all the `ZCORN` adjustments correctly applied - they are currently lost in output.

jokva:
And as usual @joakim-hove articulated the pipeline one million times better than I could.

Anywho, for those late to the party: ^ the above is our long-term goal, and this patch is hopefully a step on our way there.

PR #41, opened 2016-06-14
Inform when creating directories.

jokva:
Fairly straight forward patch - this used to be a std::cout message, now it's handled by opm-log


qilicun:
Do you think this is a good idea to introduce `OpmLog` into opm-output?

jokva:
Why wouldn't it be?

qilicun:
The reason probably is opm-output enable multi-thread but opmLog not. I'm not sure how this will influence the performance. Another reason could be @alfbr wants opm-output as a stand-alone package, like opm-parser, which I'm not sure.

jokva:
opm-output is **far** from being multi-threaded because of ert anyway. And if that really is a problem we should consider fixing opm-log, not opt-in in other modules.

I can guarantee you that the logging we might do is **dwarfed** by the work and wait needed to do disk IO, so the performance hit from logging is negligible.

atgeirr:
I think it is good to use the logging facility in opm-output.

alfbr:
Yes, I think it would be beneficial to have opm-output as an independent module, without dependencies into other repositories and with an API easily used by third parties. I believe that would increase value and interest in the module greatly. However, we are very far from that today, and that is a struggle I am not ready to embark on. Hence, feel free to use logging in opm-output.

jokva:
If that's the goal then opm-log should be moved to opm-core and opm-common should be a build-system repo only.

atgeirr:
I think opm-common should grow rather than shrink and _not_ just be a buildsystem repo (we left that behind with the opm-cmake to opm-common transition), I do not understand why you'd want that again? Also, why was this closed?

akva2:
my .02c; letting windows users' fear of dependencies run the show is pure destructive. it is not fair to let other platforms suffer just because one is limited. this is an internal dependency where opm depends on .. opm. it shouldn't even be up for discussion.

alfbr:
or we could implement a message container similar to parser, but that can be done later anyway. (please do not let this end in a windows discussion, because that is really secondary)
I did not mean for this to be closed, please review and merge at will.

jokva:
Because when half the modules can't use features because reasons it's obviously bad design.

alfbr:
There is something wrong with your count. We currently have nine modules. Only two of them are obvious candidates for broad usage stand-alone. The efforts to enable it with opm-parser was all things considered minimal, and I believe the design is very good.

jokva:
I honestly disagree.

pgdr:
This discussion has been moved to the mailing list.  This is a 1 line PR.  Focus on the review.

alfbr:
Please make the message more verbose and explicit. Something like:
"Creating directory where all output files will be written: " + outputDir
EDIT: removed comment on info, since it's already there.

PR #42, opened 2016-06-15
Fix compile and warning

atgeirr:
The compile fix is due to the same problem as was addressed in #35.

The quoted part from the standard in that PR, which version of C++ does that apply to? Certainly my compiler does not like the member syntax there.


jokva:
It's **definitely** in 14. I can look for a copy of 11 and double check.

jokva:
It's also in 11.

atgeirr:
OK, it's a compiler problem then. I hope you'll indulge my compiler and merge this... (Also I think it's clearer.)

pgdr:
I've reviewed it and it looks good, and if it fixes compile issues, it seems worth it.

PR #43, opened 2016-06-16
Well shared_ptr alias deprecated

jokva:
The shared_ptr aliasing is deprecated, and in the Well case this has
been removed upstream.

Upstream: https://github.com/OPM/opm-parser/pull/849


bska:
jenkins build this opm-parser=849 please

bska:
jenkins build this opm-parser=849 please

PR #44, opened 2016-06-16
Fix inappropriate use of EclipseGrid.

atgeirr:
The modified (revised actnum and nnc included) version should be used by the output facility instead of the unchanged input grid provided by the EclipseState.

This fixes problems with segfaults caused by the number of active cells used by the simulator not being the same as that claimed by the EclipseState::getInputGrid().

This is not a very elegant solution, but it is not very complex and I think it is a sufficient critical error to warrant an immediate fix.


atgeirr:
jenkins build this please

jokva:
This solves the wrong problem. The simulator should be creating this grid, not output, and output should **definitely** not return a grid back to the simulator.

atgeirr:
> The simulator should be creating this grid

The simulator has no need or want to manipulate EclipseGrid instances, they are a very particular thing that is only used for I/O.

> output should definitely not return a grid back to the simulator

No, and the proposal does not do it either.

jokva:
My bad, `writeHeader` is internal to INIT, but that is going to change.

jokva:
Regardless, the point still stands. I don't think this kind of logic belongs here at all, and the entire approach is subtly broken. If the simulator modifies the grid then it's up to the simulator to let the output know, not implicitly update `EclipseState`-derived structures.

atgeirr:
> If the simulator modifies the grid then it's up to the simulator to let the output know

While I can agree that this PR is a band-aid more than a cure, the patient is bleeding. If this PR is unacceptable, and there is no cure coming soon, then we should revert #30 and all that depends on it.

jokva:
This will be adressed anyway, but in its current form this won't be merged. It does not belong in `INIT` _at all_, and leads to way too many issues. I can fix it fairly fast by moving the grid to the constructor, or you can do it if you have the time.

I recommend not using `shared_ptr` for it though, opm-output is almost sane in that regard.

atgeirr:
> I can fix it fairly fast by moving the grid to the constructor

Please do.

PR #46, opened 2016-06-16
Copy and save a Grid copy in Writer; apply actnum

jokva:
Simulators might modify the grid post EclipseState creation, so the Grid
fetched from there is unreliable. Copy the Deck-provided grid and apply
the manipulations at EclipseWriter construction time to ensure it uses
the same dimensions and has the same properties as the simulator.

### DIFF
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -277,50 +277,25 @@ public:
                       const int* compressedToCartesianCellIdx,
                       time_t current_posix_time,
                       const EclipseState& es,
-                      int ert_phase_mask,
-                      const NNC& nnc = NNC() )
+                      const EclipseGrid& eclGrid,
+                      int ert_phase_mask )
     {
         auto dataField = es.get3DProperties().getDoubleGridProperty("PORO").getData();
         restrictAndReorderToActiveCells( dataField, numCells, compressedToCartesianCellIdx );
-
-        auto eclGrid = es.getInputGridCopy();
-
-        // update the ACTNUM array using the processed cornerpoint grid
-        std::vector<int> actnumData( eclGrid->getCartesianSize(), 1 );
-        if ( compressedToCartesianCellIdx ) {
-            actnumData.assign( actnumData.size(), 0 );
-
-            for( int cellIdx = 0; cellIdx < numCells; ++cellIdx ) {
-                actnumData[ compressedToCartesianCellIdx[ cellIdx ] ] = 1;
-            }
-        }
-
-        eclGrid->resetACTNUM( &actnumData[0] );
-
-        if (nnc.hasNNC())
-        {
-            int idx = 0;
-            // const_cast is safe, since this is a copy of the input grid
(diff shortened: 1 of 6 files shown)


bska:
jenkins build this please

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -548,10 +530,10 @@ void EclipseWriter::writeInit( const NNC& nnc ) {
         fortio.writeKeyword("PERMZ", data);
     }
 
-    if( !nnc.hasNNC() ) return;
+    if( !this->impl->nnc.hasNNC() ) return;
 
     std::vector<double> tran;
-    for( NNCdata nd : nnc.nncdata() ) {
+    for( NNCdata nd : this->impl->nnc.nncdata() ) {
```
blattms, line 536: I know that this has nothing to do with this PR, but it came to my attention here:
Wouldn't it be reasonable to not rely on the compiler to remove the unnecessary copy of NNCData (Two size_t and one double) and use a reference?

``` C++
for( NNCdata& nd : this->impl->nnc.nncdata() ) {
```

It would also not hurt to initialize the container with the correct size beforehand to prevent reallocations:

``` C++
std::vector<double> tran(this->impl->nnc.nncdata().size());
```

Then one could also could use some old school STL algorithms that tend to be forgotten:

``` C++
std::transform(this->impl->nnc.nncdata().begin(), this->impl->nnc.nncdata().size().end(), tran.begin(),
                       [](const NNCData& i){return nnc.trans;}); 
```

blattms:
If the simulators do the modifications is it reasonable to account for it here and introduce another copy? Or shouldn't they do the copy, save the old state and pass it wherever needed?

opm/output/eclipse/EclipseWriter.cpp
jokva, line 536: Maybe, but I sincerely doubt it will be faster. Since it's more-or-less immediately pushed back into the vector the temporary might be elided (and if not it's in super hot cache anyway). It's also 24 bytes so very unlikely to make a difference.

That being said, the reallocation thing is worse. I'm pretty sure most of this will be adapted to a heavy mapping bias once we get Windows CI to work and this file as a whole gets some love. So far it's mostly been a port from the old system modulo its opm-core dependencies. Cleanups are coming though.

jokva:
Simulator people may know otherwise, but I'm pretty sure the NNCs can be copied. They were passed and used for `writeInit` only and really just setup, so once the constructor is done (or writeInit before) it will probably not observe any changes.

It would be a nice change I suppose, but will probably be put into on hold until simulator people can confirm and I get the chance to show overall approach some love.

atgeirr:
This does not solve the segfault problem. I'll try to see why.

jokva:
Probably just me being a moron missing a couple of things. Give me a few minutes.

jokva:
Try now.

atgeirr:
Seems to work now. Will report when the run is finished.

atgeirr:
Successfully completed run on Norne without segfault.

joakim-hove:
Why does Travis fail on this?

jokva:
Corresponding PR in opm-simulator isn't merged.

PR #50, opened 2016-06-23
Prepare to write TRAN to init file.

joakim-hove:
Minor refactoring & cleanup of the writing of INIT files.

### DIFF
--- a/opm/output/eclipse/EclipseWriter.hpp
+++ b/opm/output/eclipse/EclipseWriter.hpp
@@ -51,6 +51,7 @@ public:
                   const int* compressedToCartesianCellIdx,
                   const NNC& );
 
+
     /**
      * Write the static eclipse data (grid, PVT curves, etc) to disk.
      *
@@ -79,6 +80,9 @@ public:
     ~EclipseWriter();
 
 private:
+    void writeINITFile(const std::vector<data::CellData>& simProps) const;
+    void writeEGRIDFile( ) const;
+
     class Impl;
     std::unique_ptr< Impl > impl;
 
--- a/opm/output/eclipse/EclipseWriter.hpp
+++ b/opm/output/eclipse/EclipseWriter.hpp
@@ -57,7 +57,7 @@ public:
      *
      * If NNC is given to the constructor, writes TRANNNC keyword.
      */
-    void writeInit(const std::vector<data::CellData>& simProps = {});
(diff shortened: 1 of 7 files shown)


tests/test_EclipseWriter.cpp
```diff
@@ -262,27 +264,35 @@ BOOST_AUTO_TEST_CASE(EclipseWriterIntegration)
...
-        eclWriter.writeInit();
+        std::vector<double> tranx(3*3*3);
+        std::vector<double> trany(3*3*3);
+        std::vector<double> tranz(3*3*3);
+        std::vector<data::CellData> simProps{{"TRANX" , UnitSystem::measure::transmissibility, tranx},
+                                             {"TRANY" , UnitSystem::measure::transmissibility, trany},
+                                             {"TRANZ" , UnitSystem::measure::transmissibility, tranz}};
+
+        ta.setStore( true );
+        std::cout << "Working in: " << ta.getCwd() << std::endl;
```
pgdr, line 280: debug or production?

opm/output/eclipse/EclipseWriter.hpp
```diff
@@ -51,12 +51,13 @@ class EclipseWriter {
...
                   const NNC& );
 
+
     /**
      * Write the static eclipse data (grid, PVT curves, etc) to disk.
      *
      * If NNC is given to the constructor, writes TRANNNC keyword.
      */
-    void writeInit();
+    void writeInit(const std::vector<data::CellData>& simProps = {}, const NNC& nnc = NNC());
```
pgdr, line 60: Could you add a comment also on the use of `simProps`?

pgdr:
I am ready to merge this when the minor comments are addressed.

joakim-hove:
Hmmm - I ended up doing the fixup here instead: https://github.com/OPM/opm-output/pull/51

jokva:
`writeINITFile` and `writeEGRIDFile` doesn't modify `EclipseWriter` at all, so I'd personally prefer for them to be static'd in the source file, rather than being exposed to the header. Makes the header slightly simpler, exposes fewer details (especially since we already use private implementation here) and makes its actual input-dependencies obvious via arguments.

Other than that it looks good.

joakim-hove:
I'm closing this; discussion should continue here: #51

PR #51, opened 2016-06-24
Rename write init

joakim-hove:
This is a sequel to: #50 The main work/change is in #50 - and that can be merged without downstream changes.

This PR contains some name and argument changes, and must be merged together with downstream PR: https://github.com/OPM/opm-simulators/pull/738

NB: This requires a bleeding edge ERT with: https://github.com/Ensembles/ert/pull/1133

### DIFF
--- a/opm/output/Cells.hpp
+++ b/opm/output/Cells.hpp
@@ -23,11 +23,20 @@
 #include <map>
 #include <vector>
 
+#include <opm/parser/eclipse/Units/UnitSystem.hpp>
+
 namespace Opm {
 
 class SimulationDataContainer;
 
 namespace data {
+
+    struct CellData {
+        std::string name;
+        UnitSystem::measure dim;
+        std::vector<double> data;
+    };
+
 
     struct Solution {
         /* data::Solution supports writing only some information,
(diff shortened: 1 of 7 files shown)


joakim-hove:
> @jokva: writeINITFile and writeEGRIDFile doesn't modify EclipseWriter at all, so I'd personally prefer for them to be static'd in the source file, rather than being exposed to the header. Makes the header slightly simpler, exposes fewer details (especially since we already use private implementation here)

Well - the current implementation is one small step away from the private implementation pattern; which I am not that fond of. So - I would prefer to keep things the way they are.

jokva:
Private implementation or not, they don't have to be class members (and if they don't have to it's really nice if they're not).

joakim-hove:
Well they access the EclipseState member.

jokva:
Which could just as easily be a parameter.

joakim-hove:
> Which could just as easily be a parameter.

Of course it could - but this _is_ C++ after all; I was about to give in on this but then I realized there several additional members which were accessed; so no these will remain ordinary private methods for now.

jokva:
That's my entire point really - now you're using the class to not type parameters, making it non-obvious what a function **actually** depends on without reading the entire thing. Which is ok when it relies on internal state _change_, but now they don't, they're just helpers. It's rather leaky.

joakim-hove:
See: https://github.com/OPM/opm-simulators/issues/746#issuecomment-229472143 the last commit in this PR is a band-aid fix for this.

PR #53, opened 2016-06-28
New private Summary implementation

jokva:
A rewritten Summary.cpp with some minor header modifications. Synposis
of the new implementation:
- Uses `unordered_map< string, std::function >` for dispatch, instead of
  multiple functions and a switch
- Some poor man's function composition support has been added
  (privately) to avoid a lot of reptition in the post processing.
- Functions assume they work over lists of wells instead of single wells
  being special cased - this means groups of well etc. can share
  implementation with single wells and field keywords.
- Unsupported keywords are not written in the Summary file.

Furthermore, some comments on special cases and overall approach and
a generally more declarative implementation. This change is invisible to
downstream developers. Users will obviously see no more garbage
keywords.

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -29,223 +29,113 @@
 #include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
 
+/*
+ * This class takes simulator state and parser-provided information and
+ * orchestrates ert to write simulation results as requested by the SUMMARY
+ * section in eclipse. The implementation is somewhat compact as a lot of the
+ * requested output may be similar-but-not-quite. Through various techniques
+ * the compiler writes a lot of this code for us.
+ */
+
 namespace Opm {
 
 namespace {
 
-/*
- * A series of helper functions to read & compute values from the simulator,
- * intended to clean up the keyword -> action mapping in the *_keyword
- * functions.
- */
-
 using rt = data::Rates::opt;
-
-/* The supported Eclipse keywords */
-enum class E : out::Summary::kwtype {
-    WBHP,
-    WBHPH,
-    WGIR,
-    WGIRH,
-    WGIT,
-    WGITH,
-    WGOR,
-    WGORH,
-    WGLR,
-    WGLRH,
-    WGPR,
-    WGPRH,
-    WGPT,
-    WGPTH,
-    WLPR,
-    WLPRH,
-    WLPT,
-    WLPTH,
-    WOIR,
-    WOIRH,
-    WOIT,
-    WOITH,
-    WOPR,
-    WOPRH,
-    WOPT,
-    WOPTH,
-    WTHP,
-    WTHPH,
-    WWCT,
-    WWCTH,
-    WWIR,
-    WWIRH,
-    WWIT,
-    WWITH,
-    WWPR,
-    WWPRH,
-    WWPT,
-    WWPTH,
-    GWPT,
-    GOPT,
-    GGPT,
-    GWPR,
-    GOPR,
-    GLPR,
-    GGPR,
-    GWIR,
-    GWIT,
-    GGIR,
-    GGIT,
-    GGIRH,
-    GWIRH,
-    GOIRH,
-    GGITH,
-    GWITH,
-    GWPRH,
-    GOPRH,
-    GGPRH,
-    GLPRH,
-    GWPTH,
-    GOPTH,
-    GGPTH,
-    GLPTH,
-    GWCT,
-    GGOR,
-    UNSUPPORTED,
-};
-
-const std::map< std::string, E > keyhash = {
-    { "WBHP",  E::WBHP },
-    { "WBHPH", E::WBHPH },
-    { "WGIR",  E::WGIR },
-    { "WGIRH", E::WGIRH },
-    { "WGIT",  E::WGIT },
-    { "WGITH", E::WGITH },
-    { "WGOR",  E::WGOR },
-    { "WGORH", E::WGORH },
-    { "WGLR",  E::WGLR },
-    { "WGLRH", E::WGLRH },
-    { "WGPR",  E::WGPR },
-    { "WGPRH", E::WGPRH },
-    { "WGPT",  E::WGPT },
-    { "WGPTH", E::WGPTH },
-    { "WLPR",  E::WLPR },
-    { "WLPRH", E::WLPRH },
-    { "WLPT",  E::WLPT },
-    { "WLPTH", E::WLPTH },
-    { "WOIR",  E::WOIR },
-    { "WOIRH", E::WOIRH },
-    { "WOIT",  E::WOIT },
-    { "WOITH", E::WOITH },
-    { "WOPR",  E::WOPR },
-    { "WOPRH", E::WOPRH },
-    { "WOPT",  E::WOPT },
-    { "WOPTH", E::WOPTH },
-    { "WTHP",  E::WTHP },
-    { "WTHPH", E::WTHPH },
-    { "WWCT",  E::WWCT },
-    { "WWCTH", E::WWCTH },
-    { "WWIR",  E::WWIR },
-    { "WWIRH", E::WWIRH },
-    { "WWIT",  E::WWIT },
-    { "WWITH", E::WWITH },
-    { "WWPR",  E::WWPR },
-    { "WWPRH", E::WWPRH },
-    { "WWPT",  E::WWPT },
-    { "WWPTH", E::WWPTH },
-    { "GWPT",  E::GWPT },
-    { "GOPT",  E::GOPT },
-    { "GGPT",  E::GGPT },
-    { "GWPR",  E::GWPR },
-    { "GOPR",  E::GOPR },
-    { "GLPR",  E::GLPR },
-    { "GGPR",  E::GGPR },
-    { "GWIR",  E::GWIR },
-    { "GWIT",  E::GWIT },
-    { "GGIR",  E::GGIR },
-    { "GWPRH", E::GWPRH },
-    { "GOPRH", E::GOPRH },
-    { "GGPRH", E::GGPRH },
-    { "GLPRH", E::GLPRH },
-    { "GWPTH", E::GWPTH },
-    { "GOPTH", E::GOPTH },
-    { "GGPTH", E::GGPTH },
-    { "GLPTH", E::GLPTH },
-    { "GGIRH", E::GGIRH },
-    { "GWIRH", E::GWIRH },
-    { "GOIRH", E::GOIRH },
-    { "GGIT",  E::GGIT },
-    { "GWITH", E::GWITH },
-    { "GGITH", E::GGITH },
-    { "GWCT",  E::GWCT },
-    { "GGOR",  E::GGOR },
-};
-
-inline E khash( const char* key ) {
-    /* Since a switch is used to determine the proper computation from the
-     * input node, but keywords are stored as strings, we need a string -> enum
-     * mapping for keywords.
-     */
-    auto itr = keyhash.find( key );
-    if( itr == keyhash.end() ) return E::UNSUPPORTED;
-    return itr->second;
-}
-
-inline double wct( double wat, double oil ) {
-    /* handle div-by-zero - if this well is shut, all production rates will be
-     * zero and there is no cut (i.e. zero). */
-    if( oil == 0 ) return 0;
-
-    return wat / ( wat + oil );
-}
-
-inline double wwcth( const Well& w, size_t ts ) {
-    /* From our point of view, injectors don't have meaningful water cuts. */
-    if( w.isInjector( ts ) ) return 0;
-
-    const auto& p = w.getProductionProperties( ts );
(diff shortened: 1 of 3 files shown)


jokva:
Depends on https://github.com/OPM/opm-parser/pull/862 and https://github.com/Ensembles/ert/pull/1148 which can both be merged independently.

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
+/* Some numerical value with its unit tag embedded to enable caller to apply
+ * unit conversion. This removes a ton of boilerplate.
+ *
+ * Some arithmetic operations are supported, which also codifies the following
+ * assumptions:
+ * * Division by zero, such as in gas/oil ratios without oil, shall default to 0
+ * * Division typically happen to calculate ratios, in which case unit
+ *   conversion is a no-op
+ * * When performing an operation on two operands, the left operand's unit is
+ *   applied, meaning arithmetic is **not** associative with respect to units.
```
pgdr, line 56: 1.  It should maybe be associative, no?  Maybe I'm wrong, but also maybe more importantly it's not _commutative_.
2.  Why is `x/0 = 0`?  Is this Eclipse behavoir?

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
+/*
+ * All functions must have the same parameters, so they're gathered in a struct
+ * and functions use whatever information they care about.
+ */
+struct fn_args {
+    const std::vector< const Well* >& ecl_wells;
+    double duration;
+    size_t timestep;
+    const data::Wells& wells;
+};
```
pgdr, line 88: Is `fn_args` really the best name here?

opm/output/eclipse/Summary.cpp
jokva, line 56: 1. The arithmetic is, it's the type information that behaves possibly unexpectedly. Our unit system is not complete enough to have reasonable type arithmetic (I really wish we had, see `boost-units` for a cool example), but even then it would be quirky.

Consider WGOR, well gas-oil ratio, that I'm pretty sure use different units (!!) for rates. The computation is G/O, but the result is a ratio and unitless, but unless `/` erases that then unit conversion will be applied later. Another nice one is `duration * rate` which results in volume. This **could** have been handled by unit-aware multiplication, but isn't as of now because we don't have any resemblance of support. This system takes us somewhat closer, but it does unfortunately introduce this detail.
1. I honestly don't know what eclipse does, but x/0 surely does happen naturally and returning 0 is probably better than arithmetic exceptions. Example is WGOR as above, when a well produce no oil (or switch to an injector). Regardless, the output would be garbage.

opm/output/eclipse/Summary.cpp
jokva, line 88: I personally don't think it's a bad name, but I'm open to better alternatives.

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
+ *
+ * Some arithmetic operations are supported, which also codifies the following
+ * assumptions:
+ * * Division by zero, such as in gas/oil ratios without oil, shall default to 0
+ * * Division typically happen to calculate ratios, in which case unit
+ *   conversion is a no-op
+ * * When performing an operation on two operands, the left operand's unit is
+ *   applied, meaning arithmetic is **not** associative with respect to units.
+ */
+struct uv {
```
pgdr: What about calling `uv` quantity?

>  A physical quantity can be expressed as the combination of a number – usually a real number – and a unit or combination of units

jokva: Reasonable, I'm alright with this.

tests/test_Summary.cpp
```diff
@@ -302,9 +302,9 @@ BOOST_AUTO_TEST_CASE(group_keywords) {
     BOOST_CHECK_CLOSE( 2 * 30.2, ecl_sum_get_group_var( resp, 2, "G_2", "GGIT" ), 1e-5 );
 
     /* Injection totals (history) */
-    BOOST_CHECK_CLOSE( 0,    ecl_sum_get_group_var( resp, 1, "G_2", "GWITH" ), 1e-5 );
+    BOOST_CHECK_CLOSE( 30.0, ecl_sum_get_group_var( resp, 1, "G_2", "GWITH" ), 1e-5 );
     BOOST_CHECK_CLOSE( 0,    ecl_sum_get_group_var( resp, 1, "G_2", "GGITH" ), 1e-5 );
-    BOOST_CHECK_CLOSE( 30.0, ecl_sum_get_group_var( resp, 2, "G_2", "GWITH" ), 1e-5 );
+    BOOST_CHECK_CLOSE( 60.0, ecl_sum_get_group_var( resp, 2, "G_2", "GWITH" ), 1e-5 );
```
pgdr, line 307: Why is this going from 30 to 60?

jokva, line 307: Accumulated volume.

atgeirr:
Disclaimer: just saw some of the discussion, not read the code.

About units and arithmetics: I believe that any time you need to calculate something (such as a ratio) that should always be done in SI units. Any conversion to/from other units should be done only upon (ultimate) output.

pgdr:
This is a good contribution that I'm happy to merge when the questions above are addressed and @atgeirr comment on SI unit conversion is addressed.

However, do to the fact that this is a non-trivial change, maybe we should wait some days to give others the opportunity of commenting.

jokva:
@atgeirr We are. All computation is done in SI units and it's the final result that's unit converted. I (ab)use multiplication with unit carry to "cast" the unit into identity (i.e. no unit).

@pgdr I'd like more people to look at it too. If someone could also trigger travis to run norne just to be sure that'd be nice.

tests/test_Summary.cpp
jokva, line 307: With regard to the change itself - I'm pretty sure the previous values were bugs and the current implementation seems to match Eclipse's output and are reasonably close to the simulated values (i.e. history is within reason of simulated). It's also consistent with the values for production accumulatives.

atgeirr:
> I (ab)use multiplication with unit carry to "cast" the unit into identity (i.e. no unit).

Sorry I did not understand that? (Although it is good to know that any calculations are still SI).

jokva:
> Sorry I did not understand that? (Although it is good to know that any calculations are still SI).

Consider WGOR. This is a ratio with no unit. In order to _remove_ the unit from `gas / oil`, which would yield gas flow rate, I compute `(1 identity) * ( gas / oil )` which doesn't alter the numerical value, but will apply `1`'s unit to the result. Adding `0` would accomplish the same.

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
-    { "GOIRH", E::GOIRH },
-    { "GGIT",  E::GGIT },
-    { "GWITH", E::GWITH },
-    { "GGITH", E::GGITH },
-    { "GWCT",  E::GWCT },
-    { "GGOR",  E::GGOR },
+    uv operator+( const uv& rhs ) const {
+        return { this->value + rhs.value, this->unit };
+    }
+    uv operator*( const uv& rhs ) const {
```
joakim-hove: Please - a blank line between the operators?

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
-    /* handle div-by-zero - if this well is shut, all production rates will be
-     * zero and there is no cut (i.e. zero). */
-    if( oil == 0 ) return 0;
-
-    return wat / ( wat + oil );
-}
+/*
+ * All functions must have the same parameters, so they're gathered in a struct
+ * and functions use whatever information they care about.
+ */
```
joakim-hove, line 82: Maybe two words about the difference between `ecl_wells` and `wells`?

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
 
-inline double wwcth( const Well& w, size_t ts ) {
-    /* From our point of view, injectors don't have meaningful water cuts. */
-    if( w.isInjector( ts ) ) return 0;
+/* Since there are several enums in opm scattered about more-or-less
+ * representing the same thing. Since functions use template parameters to
+ * expand into the actual implementations we need a static way to determine
+ * what unit to tag the return value with.
+ */
+template< rt > constexpr
```
joakim-hove, line 95: Again - we don't pay the rent by lines - some space would increase readability.

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
-     * guess.  If other semantics when just gas is produced, this must be
-     * changed.
-     */
-    if( liq == 0 ) return 0;
+inline uv bhp( const fn_args& args ) {
+    assert( args.ecl_wells.size() == 1 );
 
-    return gas / liq;
+    auto p = args.wells.wells.find( args.ecl_wells.front()->name() );
+    if( p == args.wells.wells.end() ) return { 0, measure::pressure };
```
joakim-hove: If we are writing for clarity - I think this is much easier on the eye:

``` C++
 if (p == arg.wells.wells.end() )
    return {0 , measure::pressure };

 return { p->second.bhp , measure::pressure };
```

joakim-hove:
All in all I think this is fine; would appreciate some reformatting for readability - but apart from that I am ready to merge this.

atgeirr:
> Consider WGOR. This is a ratio with no unit. 

That is not quite precise, it is dimensionless, but can have a unit. For example I believe in FIELD units WGOR is given/output in 1000 ft^3 / stb.

> In order to remove the unit from gas / oil, which would yield gas flow rate,
>  I compute (1 identity) \* ( gas / oil ) which doesn't alter the numerical value, but will apply 1's unit to the result. Adding 0 would accomplish the same

I just don't get this... Example in the code?

jokva:
> I just don't get this... Example in the code?

https://github.com/OPM/opm-output/pull/53/files#diff-71e13cd4acdd582de8151509e6afa2faR199

jokva:
> That is not quite precise, it is dimensionless, but can have a unit. For example I believe in FIELD units WGOR is given/output in 1000 ft^3 / stb.

But then wouldn't both oil and gas rates be in ft³ / stb? As in we wouldn't need to apply some more unit conversions on it later.

atgeirr:
> But then wouldn't both oil and gas rates be in ft³ / stb?

No, oil rate in stb/day, gas rate in 1000 ft^3/day.

alfbr:
> About units and arithmetics: I believe that any time you need to calculate something (such as a ratio) that should always be done in SI units.

Strongly supported, many of us had to learn the hard way. It is simply amazing how easy it is to get something wrong in unit conversions.

opm/output/eclipse/Summary.cpp
```diff
@@ -29,223 +29,103 @@
...
-    { "GLPTH", E::GLPTH },
-    { "GGIRH", E::GGIRH },
-    { "GWIRH", E::GWIRH },
-    { "GOIRH", E::GOIRH },
-    { "GGIT",  E::GGIT },
-    { "GWITH", E::GWITH },
-    { "GGITH", E::GGITH },
-    { "GWCT",  E::GWCT },
-    { "GGOR",  E::GGOR },
+    uv operator+( const uv& rhs ) const {
```
atgeirr: All of these operators are in my opinion misleading and dangerous. Addition of quantities with different units should not be allowed, that is silently accepted and the lhs unit used for the result. Multiplication should yield a unit equal to the product unit, not the lhs unit. Division should yield the quotient unit, not identity.

jokva: They're also highly internal (which is why I think this behaviour is ok-ish because they solve a real problem) and unit support in opm is atrocious. I'd be **very** happy to implement or adopt a proper unit system which would solve all this, but as of now it's not feasible at all.

atgeirr:
> unit support in opm is atrocious

I think that is a bit overblown, what is atrocious about it? I agree that there is some duplication between opm-parser and opm-core (candidate for opm-common in my opinion) which is unfortunate, but I guess that is not what bothers you?

jokva:
> I think that is a bit overblown, what is atrocious about it?

For one, it doesn't exist. Nothing but doubles are passed around and we pray that enough unit information is passed around and that everyone remembers to check and apply conversions only once. Which is fine inside the simulator where you can presume everything uses SI with the right factors, but that doesn't apply nor carry in communication.

`UnitSystem` and friends is **not** unit support.

atgeirr:
Whenever doubles are passed around, they _are_ SI. The only exception should be the input and output facilities, and there you must of course make sure you separate SI from non-SI. That you must do by clear naming of functions and variables so that it is never in doubt to the reader/maintainer. It seems to me that you do not consider anything as unit support unless it somehow embeds unit information with values? That is only one way to deal with things. The best way? I doubt it. I think the ideal (that we should perhaps consider moving towards) is to embed (compile time) dimension information with values. That would give us compile time dimension checking at no runtime cost. Converting to and from SI should still happen on the I/O boundary.

jokva:
That's what I'm sort-of suggestion (boost-units was mentioned earlier), but that's a massive task. The outcome of it would trivially solve this, but as long as we have virtually nothing it doesn't matter.

PR #55, opened 2016-06-30
C++ regression test of eclipse files (initial, unified restart and summary)

ptaule:
This is essentially the same pull request as OPM/opm-common#153, but in a different module. The code compares eclipse files (.INIT, .UNSRT, .UNSMRY) written by flow or ECLIPSE. Note that the final version of this PR includes #56.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -31,6 +31,20 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/test_util/summaryComparator.hpp
     )
 
+list (APPEND EXAMPLE_SOURCE_FILES
+        test_util/initRegressionTest.cpp
+        test_util/restartRegressionTest.cpp
+        test_util/summaryRegressionTest.cpp
+    )
+
+# programs listed here will not only be compiled, but also marked for
+# installation
+list (APPEND PROGRAM_SOURCE_FILES
+        test_util/initRegressionTest.cpp
+        test_util/restartRegressionTest.cpp
+        test_util/summaryRegressionTest.cpp
+    )
+
 list (APPEND TEST_SOURCE_FILES
         tests/test_compareSummary.cpp
         tests/test_EclFilesComparator.cpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -8,6 +8,9 @@ list( APPEND MAIN_SOURCE_FILES
         opm/output/eclipse/Summary.cpp
         opm/output/eclipse/writeECLData.cpp
         opm/output/vtk/writeVtkData.cpp
+        opm/test_util/EclFilesComparator.cpp
+        opm/test_util/summaryRegressionTest.cpp
+        opm/test_util/summaryComparator.cpp
     )
 
 list (APPEND PUBLIC_HEADER_FILES
@@ -23,22 +26,27 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/output/eclipse/Summary.hpp
         opm/output/eclipse/writeECLData.hpp
         opm/output/vtk/writeVtkData.hpp
+        opm/test_util/EclFilesComparator.hpp
+        opm/test_util/summaryRegressionTest.hpp
+        opm/test_util/summaryComparator.hpp
     )
 
 list (APPEND TEST_SOURCE_FILES
+        tests/test_compareSummary.cpp
+        tests/test_EclFilesComparator.cpp
         tests/test_EclipseWriter.cpp
-        tests/test_RFT.cpp
-        tests/test_writenumwells.cpp
         tests/test_Restart.cpp
-        tests/test_Wells.cpp
+        tests/test_RFT.cpp
         tests/test_Summary.cpp
-)
+        tests/test_Wells.cpp
+        tests/test_writenumwells.cpp
+    )
 
 # originally generated with the command:
 # find tests -name '*.xml' -a ! -wholename '*/not-unit/*' -printf '\t%p\n' | sort
 list (APPEND TEST_DATA_FILES
-        tests/testBlackoilState3.DATA
-        tests/summary_deck.DATA
         tests/FIRST_SIM.DATA
+        tests/summary_deck.DATA
+        tests/testBlackoilState3.DATA
         tests/testRFT.DATA
-)
+    )
--- a/opm/test_util/EclFilesComparator.cpp
+++ b/opm/test_util/EclFilesComparator.cpp
@@ -1,0 +1,305 @@
+/*
+   Copyright 2016 Statoil ASA.
+
+   This file is part of the Open Porous Media project (OPM).
+
+   OPM is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   OPM is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+   */
+
+#include <opm/test_util/EclFilesComparator.hpp>
+#include <opm/common/ErrorMacros.hpp>
+
+#include <iostream>
+#include <algorithm>
+#include <ert/ecl/ecl_file.h>
+#include <ert/ecl/ecl_grid.h>
+
+bool ECLFilesComparator::keywordValidForComparing(const std::string& keyword) const {
+    auto it = std::find(keywords1.begin(), keywords1.end(), keyword);
+    if (it == keywords1.end()) {
+        std::cout << "Could not find keyword: " << keyword << std::endl;
+        OPM_THROW(std::runtime_error, "Keyword does not exist in first file.");
+    }
+    it = find(keywords2.begin(), keywords2.end(), keyword);
+    if (it == keywords2.end()) {
+        std::cout << "Could not find keyword: " << keyword << std::endl;
+        OPM_THROW(std::runtime_error, "Keyword does not exist in second file.");
+    }
+    ecl_kw_type* ecl_kw = ecl_file_iget_named_kw(ecl_file1, keyword.c_str(), 0);
+    ecl_type_enum ecl_kw_type = ecl_kw_get_type(ecl_kw);
+    switch (ecl_kw_type) {
+        case ECL_DOUBLE_TYPE:
+            return true;
+        case ECL_FLOAT_TYPE:
+            return true;
+        case ECL_CHAR_TYPE:
+            std::cout << "\nKeyword " << keyword << " is of type ECL_CHAR_TYPE, "
+                << "which is not supported for comparison." << std::endl;
+            return false;
+        case ECL_INT_TYPE:
+            std::cout << "\nKeyword " << keyword << " is of type ECL_INT_TYPE, "
+                << "which is not supported for comparison." << std::endl;
+            return false;
+        case ECL_BOOL_TYPE:
+            std::cout << "\nKeyword " << keyword << " is of type ECL_BOOL_TYPE, "
+                << "which is not supported for comparison." << std::endl;
+            return false;
+        case ECL_MESS_TYPE:
+            std::cout << "\nKeyword " << keyword << " is of type ECL_MESS_TYPE, "
+                << "which is not supported for comparison." << std::endl;
+            return false;
+        default:
+            std::cout << "\nKeyword " << keyword << "has undefined type." << std::endl;
+            return false;
+    }
+}
+
+
+
+void ECLFilesComparator::printResultsForKeyword(const std::string& keyword) const {
+    const double absDeviationAverage = average(absDeviation);
+    const double relDeviationAverage = average(relDeviation);
+    std::cout << "Average absolute deviation = " << absDeviationAverage  << std::endl;
+    std::cout << "Median absolute deviation  = " << median(absDeviation) << std::endl;
+    std::cout << "Average relative deviation = " << relDeviationAverage  << std::endl;
+    std::cout << "Median relative deviation  = " << median(relDeviation) << "\n\n";
+}
+
+
+
+void ECLFilesComparator::deviationsForOccurence(const std::string& keyword, int occurence) {
+    ecl_kw_type* ecl_kw1         = ecl_file_iget_named_kw(ecl_file1, keyword.c_str(), occurence);
+    ecl_kw_type* ecl_kw2         = ecl_file_iget_named_kw(ecl_file2, keyword.c_str(), occurence);
+    const unsigned int numCells1 = ecl_kw_get_size(ecl_kw1);
+    const unsigned int numCells2 = ecl_kw_get_size(ecl_kw2);
+    if (numCells1 != numCells2) {
+        std::cout << "For keyword " << keyword << " and occurence " << occurence << ":\n";
+        OPM_THROW(std::runtime_error, "Number of active cells are different.");
+    }
+    std::vector<double> values1(numCells1), values2(numCells2);
+    ecl_kw_get_data_as_double(ecl_kw1, values1.data());
+    ecl_kw_get_data_as_double(ecl_kw2, values2.data());
+
+    auto it = std::find(keywordWhitelist.begin(), keywordWhitelist.end(), keyword);
+    for (unsigned int cell = 0; cell < values1.size(); cell++) {
+        deviationsForCell(values1[cell], values2[cell], keyword, occurence, cell, it == keywordWhitelist.end());
+    }
+}
+
+
+
+void ECLFilesComparator::deviationsForCell(double val1, double val2, const std::string& keyword, int occurence, int cell, bool allowNegativeValues) {
+    int i, j, k;
+    ecl_grid_get_ijk1(ecl_grid1, cell, &i, &j, &k);
+    // Coordinates from this function are zero-based, hence incrementing
+    i++, j++, k++;
+
+    if (showValues) {
+        std::cout << "Occurence        = "  << occurence << std::endl;
+        std::cout << "Grid coordinate  = (" << i << ", " << j << ", " << k << ")\n";
+        std::cout << "(value1, value2) = (" << val1 << ", " << val2 << ")\n\n";
+    }
+    if (!allowNegativeValues) {
+        if (val1 < 0) {
+            if (std::abs(val1) > absTolerance) {
+                std::cout << "For keyword " << keyword << " (which is listed for positive values only), occurence " << occurence
+                    << ", and grid coordinate (" << i << ", " << j << ", " << k << "):\n"
+                    << "the vector has a value of " << val1 << ", which exceeds the absolute tolerance of " << absTolerance << ".\n";
+                OPM_THROW(std::runtime_error, "Negative value in first file.");
+            }
+            val1 = 0;
+        }
+        if (val2 < 0) {
+            if (std::abs(val2) > absTolerance) {
+                std::cout << "For keyword " << keyword << " (which is listed for positive values only), occurence " << occurence
+                    << ", and grid coordinate (" << i << ", " << j << ", " << k << "):\n"
+                    << "the vector has a value of " << val2 << ", which exceeds the absolute tolerance of " << absTolerance << ".\n";
+                OPM_THROW(std::runtime_error, "Negative value in second file.");
+            }
+            val2 = 0;
+        }
+    }
+    Deviation dev = calculateDeviations(val1, val2);
+    if (dev.abs > absTolerance && dev.rel > relTolerance) {
+        std::cout << "For keyword " << keyword << ", occurence " << occurence
+            << ", and grid coordinate (" << i << ", " << j << ", " << k << "):\n"
+            << "(first value, second value) = (" << val1 << ", " << val2 << ")\n";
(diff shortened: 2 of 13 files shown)


applications/compareRestart.cpp
```diff
@@ -0,0 +1,187 @@
...
+    if (val1 < 0) {
+        val1 = 0;
+        //std::cout << "Value 1 has a negative quantity." << std::endl; 
+    }
+    if (val2 < 0) {
+        val2 = 0;
+        //std::cout << "Value 2 has a negative quantity." << std::endl; 
+    }
+
+    if (!(val1 == 0 && val2 == 0)) {
```
akva2: if (val1 != 0 && val2 != 0)

applications/compareRestart.hpp
```diff
@@ -0,0 +1,61 @@
+/*
```
akva2: what's the purpose of this header file? i don't see a reason for us to expose application internals to the world.

akva2: i guess exposing the UNSTReader might make sense, but in that case it should go in a separate compilation unit, and only the class definition be exposed. you can then include and use this class from the application. that makes sense if you want to build more than one application using the class.

joakim-hove:
OK - slightly ironic that I am the one to merge this ;-)

Some comments:
1. I want this in a directory `test_util/` instead of `applications`.
2. How do you plan to invoke these binaries? On the one hand they must be visible/available from other modules - but I guess you do not see them as part of the general `make install` target? 
3. I would prefer if you in parallell with this development opened and maintained a PR in `opm-simulators` actually _using_ the comparison binaries.

applications/compareRestart.cpp
bska: > `if (val1 != 0 && val2 != 0)`

That's not the same condition.

akva2: indeed. ||

CMakeLists_files.cmake
```diff
@@ -8,6 +8,7 @@ list( APPEND MAIN_SOURCE_FILES
         opm/output/eclipse/Summary.cpp
         opm/output/eclipse/writeECLData.cpp
         opm/output/vtk/writeVtkData.cpp
+				applications/compareRestart.cpp
```
akva2: indent.

CMakeLists_files.cmake
```diff
@@ -25,6 +26,10 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/output/vtk/writeVtkData.hpp
     )
 
+	list (APPEND EXAMPLE_SOURCE_FILES
+				applications/compareRestart.cpp
```
akva2: indent.

applications/compareRestart.hpp
ptaule: In the C++ introduction course I took, we made header files whenever we created classes -- so I guess it is a (bad?) habit of mine. I see the point about not exposing it to the world, but I believe I want to do almost the same when comparing for example init-files, so I might want to use the class in other tests. Also, to make unit tests, won't I need to be able to include he header file to be able to test the functions in compareRestart.cpp in a unit test cppfile?

> but in that case it should go in a separate compilation unit

what does this mean, and how do I do this?

akva2: this means you create an UNSTReader.hpp with only its prototype, (i.e. this file but without the printHelp()) and an UNSTReader.cpp with the implementation. these should then be compiled into the opm-output library. finally you include UNSTReader.hpp in the test application.

usage of this in unit tests would correspond to my 'more than one application using the class'.

ptaule:
1. The directory name `applications` was @alfbr's wish :)
2. I have not given much thought to this, but if I mark it for installation as commented by @akva2 in #56, would this answer the question?
3. Like for example a script running flow, and invoking `compareRestart` to compare with earlier runs, then?

akva2:
1. honor joakim's request
2. yes. that's taken care of with the comment in #56 
3. that's precisely what's requested. see https://github.com/OPM/opm-upscaling/blob/master/tests/runtest.sh for such a script (using another comparison tool) and https://github.com/OPM/opm-upscaling/blob/master/TestUpscalingBinaries.cmake for how to add a test using opm_add_test()

applications/compareRestart.hpp
atgeirr: Also (thought this was not said already, sorry if I repeat stuff) the header and class implementation should then go in one of the "usual" opm directories, for example opm/output/eclipse, the header marked for installation in the CMakeList_files.cmake, and the program using it should include it with full path (`#include <opm/...>`).

ptaule: ^ Thanks, I was just about to ask about that!

test_util/compareRestart.cpp
```diff
@@ -0,0 +1,81 @@
...
+   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+   */
+
+#include <opm/output/eclipse/EclFilesComparator.hpp>
+
+#include <iostream>
+
+#include <ert/ecl/ecl_grid.h>
+
+void printHelp();
```
akva2: nitpicking, but you could

1) move this above main so you don't have to prototype it
2) make it static void printHelp(). a static function in a source file is hidden from the linker, so you avoid possible name conflicts (not that anybody else should have claimed printHelp() but for educational purposes..)

akva2:
starts to look very good now! note that i will not merge before usage is in place to prove that the tool does the intended job, i.e. the downstream test in opm-simulators.

CMakeLists_files.cmake
```diff
@@ -38,7 +38,7 @@ list (APPEND PROGRAM_SOURCE_FILES
 	)
 
 list (APPEND TEST_SOURCE_FILES
-	#tests/test_calculations.cpp
+	tests/test_calculations.cpp
```
akva2: i don't think the name of this is optimal. why not test_EclFilesComparator ?

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+   */
+
+#include <opm/output/eclipse/EclFilesComparator.hpp>
+
+ECLFilesComparator::ECLFilesComparator(const char* unrstFile1, const char* unrstFile2) {
+    ecl_file1 = ecl_file_open(unrstFile1, ECL_FILE_CLOSE_STREAM);
+    ecl_file2 = ecl_file_open(unrstFile2, ECL_FILE_CLOSE_STREAM);
```
joakim-hove: The  `ECL_FILE_CLOSE_STREAM` will close the `FILE *` between every keyword access, this is a weird construction to limit use of filedescriptors (mainly on windows) - no reason to use this flag; just pass `0`.

test_util/compareRestart.cpp
```diff
@@ -0,0 +1,79 @@
...
+    try {
+        const char* gridFile1 = argv[1];
+        const char* gridFile2 = argv[2];
+        const char* unrstFile1 = argv[3];
+        const char* unrstFile2 = argv[4];
+        // Comparing grid sizes from .EGRID-files
+        {
+            ecl_grid_type* ecl_grid1 = ecl_grid_alloc(gridFile1);
+            ecl_grid_type* ecl_grid2 = ecl_grid_alloc(gridFile2); 
+            std::cout << "\nName of grid1: " << ecl_grid_get_name(ecl_grid1) << std::endl;
```
joakim-hove: There is a function `ecl_grid_equal (...)` which will compare all the vertices in the two grids,  this is a quite strict test which might fail when comparing an Eclipse reference grid and a grid from `flow` when `MINPV` or similar grid modifiers are active.

ptaule:
Now I have added the possibility for comparing arbitrary vectors, and also made a compareInit.cpp which uses the same class to compare vectors in INIT-files. Some of the restart vectors, for example `LOGIHEAD`, `ZWEL`, `STARTSOL` and `ENDSOL` are not supported yet however, because it seems that they do not contain values(?)

The tools can be invoked by the shellscript and CTest in OPM/opm-simulators#754 and OPM/opm-common#158.

akva2:
great stuff. you can have a beer this weekend with a smile on your face. i'll get to this on monday!

ptaule:
I restructured the code one more time: now, when exceptions are thrown for deviations, the report step and cell coordinate in question is printed. In addition, the class now handles both init and restart files, which makes `compareRestart.cpp` and `compareInit.cpp` equal except for the file flag sent to the `ECLFilesComparator` constructor. Hence it might be possible have one cppfile with restart/init as CL argument in stead, but I left it as it is for now. 

I'd say the PR is ready for a review now, @akva2 :)

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,307 @@
...
+    ecl_kw_type* ecl_kw2               = ecl_file_iget_named_kw(ecl_file2, keyword.c_str(), reportStep);
+    const unsigned int numActiveCells1 = ecl_kw_get_size(ecl_kw1);
+    const unsigned int numActiveCells2 = ecl_kw_get_size(ecl_kw2);
+    if (numActiveCells1 != numActiveCells2) {
+        std::cout << "For keyword " << keyword << " and report step " << reportStep << ":\n";
+        OPM_THROW(std::runtime_error, "Number of active cells are different.");
+    }
+    std::vector<double> values1, values2; // Elements in the vector corresponds to active cells
+    values1.resize(numActiveCells1);
+    values2.resize(numActiveCells2);
```
akva2: use the constructor, `std::vector<double> values1(numActiveCells1), values2(numActiveCells2);`

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,307 @@
...
+    int i, j, k;
+    ecl_grid_get_ijk1A(ecl_grid1, cell, &i, &j, &k);
+
+    if (showValues) {
+        std::cout << "Report step      = " << reportStep << std::endl;
+        std::cout << "Grid coordinate  = (" << i + 1 << ", " << j + 1 << ", " << k + 1 << ")\n";
+        std::cout << "(value1, value2) = (" << val1 << ", " << val2 << ")\n\n";
+    }
+    if (val1 < 0) {
+        if (abs(val1 > absTolerance)) {
```
akva2: watch your parantheses.

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,307 @@
...
+            std::cout << "For keyword " << keyword << ", report step " << reportStep 
+                << ", and grid coordinate (" << i + 1 << ", " << j + 1 << ", " << k + 1 << "):\n";
+            OPM_THROW(std::runtime_error, "Negative value in first file with absolute value larger than absolute tolerance specified.");
+        }
+        val1 = 0;
+    }
+    if (val2 < 0) {
+        if (abs(val2 > absTolerance)) {
```
akva2: parantheses. abs(val2) > absTolerance

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+
+
+void ECLFilesComparator::printKeywords() const {
+    std::cout << "\nKeywords for first file:\n";
+    for (unsigned int i = 0; i < keywords1.size(); ++i) {
+        std::cout << keywords1[i] << std::endl;
+    }
+    std::cout << "\nKeywords for second file:\n";
+    for (unsigned int i = 0; i < keywords1.size(); ++i) {
```
akva2: keywords2.size(). also i'd do 
`for (const auto& it : keywords2)
  std::cout << it << std::endl;
`

opm/output/eclipse/EclFilesComparator.cpp
```diff
@@ -0,0 +1,307 @@
...
+    return true;
+}
+
+
+
+void ECLFilesComparator::results() {
+    if (keywords1.size() != keywords2.size()) {
+        OPM_THROW(std::runtime_error, "Number of keywords are not equal.");
+    }
+    for (unsigned int i = 0; i < keywords1.size(); ++i) {
```
akva2: i'd do a range for (see above).

opm/output/eclipse/EclFilesComparator.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+   */
+
+#ifndef ECLFILESCOMPARATOR_HPP
+#define ECLFILESCOMPARATOR_HPP 
+
+#include <vector>
+#include <opm/common/ErrorMacros.hpp>
```
akva2: not needed in header.

opm/output/eclipse/EclFilesComparator.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+struct Deviation {
+    double abs = -1;
+    double rel = -1;
+};
+
+
+
+
+class ECLFilesComparator {
```
akva2: this is a public interface. i expect doxy on at least the part exposed to the world.

akva2:
additional request: my OCD twitched. i think we should stick the library-files in an opm/test_util/ folder to keep it separate from the actual writing code.

opm/output/test_util/EclFilesComparator.cpp
```diff
@@ -0,0 +1,297 @@
...
+    std::cout << "Average absolute deviation = " << absDeviationAverage  << std::endl;
+    std::cout << "Median absolute deviation  = " << median(absDeviation) << std::endl;
+    std::cout << "Average relative deviation = " << relDeviationAverage  << std::endl;
+    std::cout << "Median relative deviation  = " << median(relDeviation) << "\n\n";
+}
+
+
+
+void ECLFilesComparator::deviationsForReportStep(const std::string& keyword, int reportStep) {
+    ecl_kw_type* ecl_kw1               = ecl_file_iget_named_kw(ecl_file1, keyword.c_str(), reportStep);
```
joakim-hove: The variable here is not really `reportStep` but rather a flat index. A unified restart file can - and quite often _will_ - contain "holes" in the series of report steps - e.g. the report steps available in the unified report from norne can be read off [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/IntegrationTests/IOConfigIntegrationTest.cpp#L53) as 0-5, 10, 19, 27, ...

The simplest "fix" is just to rename the variable to e.g. `occurence` or just plain `index`  - if you _really_ mean report step the `ecl_file` code must be updated: https://github.com/Ensembles/ert/blob/master/devel/libecl/src/ecl_rstfile.c - I would start with just renaming.

opm/output/test_util/EclFilesComparator.hpp
```diff
@@ -0,0 +1,108 @@
...
+#include <string>
+
+/// Prototype for eclipse file struct, defined in the ERT library
+struct ecl_file_struct;
+typedef struct ecl_file_struct ecl_file_type;
+/// Prototype for eclipse grid struct, defined in the ERT library
+struct ecl_grid_struct;
+typedef struct ecl_grid_struct ecl_grid_type;
+
+enum eclFileEnum {RESTARTFILE, INITFILE};
```
joakim-hove: Of course you could use this - but as long as you are using ert functions for these I would rather suggest using the [ert enum](https://github.com/Ensembles/ert/blob/master/devel/libecl/include/ert/ecl/ecl_util.h#L32)

ptaule: Yeah, I thought about that. The reason why I figured I'd create my own was that I only want restart, init, rft(?) .. and not summary files since they are covered by #56, and I would also have to include the entire ecl_util.h only to be able to use the small enum def – which seemed unnecessary.

CMakeLists_files.cmake
```diff
@@ -22,17 +25,22 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/output/eclipse/EclipseWriter.hpp
         opm/output/eclipse/Summary.hpp
         opm/output/eclipse/writeECLData.hpp
-        opm/output/vtk/writeVtkData.hpp
+				opm/output/vtk/writeVtkData.hpp
```
akva2: mip.

opm/test_util/EclFilesComparator.cpp
```diff
@@ -108,15 +108,15 @@ void ECLFilesComparator::deviationsForCell(double val1, double val2, const std::
         std::cout << "(value1, value2) = (" << val1 << ", " << val2 << ")\n\n";
     }
     if (val1 < 0) {
-        if (abs(val1) > absTolerance) {
+        if (std::abs(val1) > absTolerance) {
```
blattms: May I ask what problem the qualifier is supposed to solve?

In DUNE there have recently been [several commits](https://gitlab.dune-project.org/core/dune-grid/merge_requests/73) removing the std qualifiers. It is claimed that this has been done  to make [argument dependent lookup](https://en.wikipedia.org/wiki/Argument-dependent_name_lookup) work.

blattms: Is there maybe an `#include<cmath>` missing to make `abs(double)` work?

akva2: it's a real issue. the problem is that abs(x) resolves to the function with signature 

``` C++
int abs(int x)
```

i.e. it applies a silent implicit cast to x. this depends on which headers happen to be included. it bit here on my toolchain.

bska: > It is claimed that this has been done  to make argument dependent lookup work.

True, but then you also need to insert

``` C++
using std::abs;
```

in an appropriate enclosing scope.

blattms: @akva2 Indeed, it is the same on my machine. It always calls the int version. Weired.

For completeness, here is my testcase:

``` c++
#include<cmath>
#include<cstdlib>
#include<iostream>
int main()
{
    double d=-1.5;
    std::cout<<"abs="<<abs(d)<<std::endl;
}
```

with output `abs=1`

akva2: there is actually a system to the madness. sort of. abs() is a C function, for floats it is fabs().

however, in c++ there is std::abs which is a template. so if somebody has include <cmath> and flattened the std namespace, it resolves to that template. this is why it sometimes work as expected without you doing anything (but that's all an accident). you have three choices;

1) use fabs() - that always works but you are doing C not C++, this might bite in templates.
2) explicit qualification like here
3) using namespace std, or better using std:.abs;

i prefer 2) because there's no possibility for surprises that way (that somehow the wrong symbol get used on a stupid compiler or whatever).

the whole mess is basically a C compatibility issue. to add to the fun: visual studio always exposes the template so this is one of those watch-out-for points when i port visual studio code..

blattms: @akva2,  @bsa Learned something new again. (C++: complicated; my Mind: simple).

jokva: I'm 100% with @akva2 on this on - unless you specifically want ADL because you're already in a templated function that can take a custom `abs` for your custom number type, use explicit `std::abs`. I usually go so far as to _never_ import the `std` namespace into the global namespace and always qualify my `std` or third-party libraries.

It's unfortunate that books, articles and classes bother to do this.

jokva:
This doesn't belong in this repository.

jokva:
I've reverted this merge. Even though these files are created by output there's no reason for these applications to live here.

Feel free to resubmit the patch somewhere more approperiate.

joakim-hove:
The revert of this PR in #68 is not ok. It is an honest thing to disagree about where this code should go[1], but it has been targetting opm-output for more than three weeks now. Voicing public disagreement by reverting the already merged PR is way too late, and totally unaccaptable.

I encourage @akva2 to use the necessary git fu to recover this PR in the way he ses fit, and merge the final result.

[1]: Yes I have myself been sceptical to see this code in opm-output, personally I would have preferred opm-simulators, but the decission has been taken and reverting PR's is not the way to make your voice heard.

jokva:
I guess that makes @akva2 the maintainer of this repository.

akva2:
I think there has been a misunderstanding about the process here.

I was given the task of supervising and landing the work of Kristian and Petter.
After discussions the decision was taken to put this code in opm-output. The first thing I then did was to call @joakim-hove and clarify things. I was given the authority to merge the code when I deemed it ready. I am in no way maintainer of this repository, and I have in no way disrespected the maintainers. This was not a wild-west merge, everything had been decided upon up front.

test_util/summaryRegressionTest.cpp
```diff
@@ -0,0 +1,85 @@
+/*
+   Copyright 2016 Statoil ASA.
+   This file is part of the Open Porous Media project (OPM).
+
+   OPM is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+   char* unit = getUnit(keyword);
```
pgdr, line 9: Please fix copyright header.  That goes for the other files as well.

test_util/summaryRegressionTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+   */
+
+#include <opm/test_util/summaryRegressionTest.hpp>
+#include <string>
+#include <getopt.h>
+void printHelp(){
+    std::cout << "The program takes four arguments"          << std::endl;
+    std::cout << "1) <path to file1>/<base_name>"            << std::endl;
+    std::cout << "2) <path to file2>/<base_name>"            << std::endl;
+    std::cout << "the basename should be without ectension." << std::endl;
```
pgdr, line 26: typo

ptaule, line 26: Thanks for pointing out! This and the comment above will be fixed in #71

PR #56, opened 2016-06-30
C++ summary regression test

kristfho:
Added a test that compares two summary files.


CMakeLists_files.cmake
```diff
@@ -25,6 +25,10 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/output/vtk/writeVtkData.hpp
     )
 
+list (APPEND EXAMPLE_SOURCE_FILES
+		 applications/compareSummary.cpp
+)
+
```
akva2, line 46: this is my fault. i forgot to mention that in order for an application to be marked for installation, it also needs to be added to the PROGRAM_SOURCE_FILES variable. that is, EXAMPLE_SOURCE_FILES is for being built, PROGRAM_SOURCE_FILES is for being installed.

applications/compareSummary.cpp
```diff
@@ -0,0 +1,365 @@
...
+  (at your option) any later version.
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "compareSummary.hpp"
```
akva2: same comment i had for petter. i do not see why we need this header file, it only contains application internals. unless you plan to reuse the summary comparator class elsewhere, in that case that should go to a separate header/implementation file (which is built into the opm-output library). but there is no need and no desire to expose application internals like printHelp() to the world.

applications/compareSummary.cpp
```diff
@@ -0,0 +1,365 @@
...
+  const char * unsmryFile2 = argv[4];
+  double relative_tolerance_max = 1 ;
+  double relative_tolerance_median_max = 0.1;
+  stringlist_type * file1 = stringlist_alloc_new();
+  stringlist_type * file2 = stringlist_alloc_new();
+  stringlist_append_copy(file1, unsmryFile1);
+  stringlist_append_copy(file2, unsmryFile2);
+ 
+
+  SummaryReader read(relative_tolerance_max,relative_tolerance_median_max); //may also use sat function for tolerance limits
```
akva2: this needs to be in a try-catch block. you are throwing in the reader, you need to catch those exceptions (and report / graciously handle the failure).

applications/compareSummary.cpp
```diff
@@ -0,0 +1,460 @@
...
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
```
pgdr: this is a bit nitpicking that I don't really care that much about, but why didn't you use the default header?  See  [Cells.hpp](https://github.com/OPM/opm-output/blob/master/opm/output/Cells.hpp)

applications/compareSummary.cpp
```diff
@@ -0,0 +1,460 @@
...
+  SummaryReader(): 
+    ecl_sum1(nullptr), 
+    ecl_sum2(nullptr),
+    ecl_sum_file_short(nullptr), 
+    ecl_sum_file_long(nullptr),
+    referance_vec(nullptr),
+    ref_data_vec(nullptr),
+    checking_vec(nullptr),
+    check_data_vec(nullptr),
+     keys1(stringlist_alloc_new()),
```
pgdr: strange indent

applications/compareSummary.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+
+  bool open(const char* smspecFile1, const stringlist_type* unsmryFile1, const char* smspecFile2, const stringlist_type* unsmryFile2);
+  void close();
+  void setDataSets(std::vector<double> &time_vec1,std::vector<double> &time_vec2);
+  void setKeys(); // Creates a keylist containing all the keywords in a summary file. When the number of keywords are not
+  // equal in the two files of interest, it figures out which that contains most/less keywords
+  void setToleranceLevels(double relative_tolerance_max, double relative_tolerance_median_max);
+  void setTimeVecs(std::vector<double> &time_vec1,std::vector<double> &time_vec2);
+  void setDataVecs(std::vector<double> &data_vec1,std::vector<double> &data_vec2, int index1, int index2); //Read the data from the two files into separate vectors. Not necessary the same amount of values, but the values correspond to the same interval in time. Thus possible to interpolate values.
```
pgdr: this was a remarkably long line.  Maybe add some line spacing and break the line?  Maybe put it above the `setDataVecs` function.

applications/compareSummary.cpp
```diff
@@ -0,0 +1,460 @@
...
+ 
+
+  SummaryReader read(relative_tolerance_max,relative_tolerance_median_max); //may also use sat function for tolerance limits
+  read.open(smspecFile1, file1, smspecFile2, file2);
+  read.setKeys();
+  read.getDeviations();
+  read.close();
+  stringlist_free(file1);
+  stringlist_free(file2);
+   }
```
pgdr: this was also a strange indentation.  Could you through the entire file and fix indentation issues?

applications/compareSummary.cpp
```diff
@@ -0,0 +1,460 @@
...
+    data_vec2.push_back(ecl_sum_iget(ecl_sum_file_long, time_index, int_vector_iget( var_index2 , index2 )));
+  }
+ 
+
+   
+}
+
+
+void SummaryReader::chooseReferance(std::vector<double> &time_vec1,std::vector<double> &time_vec2,std::vector<double> &data_vec1,std::vector<double> &data_vec2){
+  //This function is suppose to choose which vector that is referance vector, but when it is in use, get segmentation error
```
pgdr: I don't understand this comment.  What do you mean "when it is in use, get segmentation error"?  When the vector is in use or the function?

applications/compareSummary.hpp
```diff
@@ -0,0 +1,86 @@
...
+  stringlist_type * keys_short;
+  stringlist_type * keys_long;
+  double relative_tolerance_max;
+  double relative_tolerance_median_max;
+  std::vector<double> * referance_vec;
+  std::vector<double> * ref_data_vec;
+  std::vector<double> * checking_vec;
+  std::vector<double> * check_data_vec;
+public:
+  SummaryReader():
```
pgdr: This constructor seems to be implemented both in the header file and source file.

kristfho:
I'm writing some unit tests, but i keep getting a linker error. Should I separate the function implementation, the function declaration and the main function, like Petter have done  here #55 ? (Separate the class and the application)

akva2:
yes please

akva2:
and the reason you get linker failures: since the code sit in the main application .cpp, it's only compiled into that application. 

when you include a header prototyping a class you tell the compiler 'this code is somewhere else and will be found during linking'. that is not the case in your unit test as-is. 

hence we stick it into the library; all tests are linked against the library, linker finds the code and thus cookie for everybody.

pgdr:
Could the linker error be caused by [constructors defined several places](https://github.com/OPM/opm-output/pull/56#discussion_r69124438)?

applications/compareSummary.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+struct Deviation {
+  double relative_deviation; 
+  double absolute_deviation; 
+};
+
+
+class SummaryReader {
+private:
+  ecl_sum_type* ecl_sum1;
```
atgeirr: I think we now can use the C++11 style inline default initializers:

``` C++
ecl_sum_type* ecl_sum1 = nullptr;
```

That cuts down on the boilerplate and on possibility for errors, too!

opm/output/eclipse/summaryReader.cpp
```diff
@@ -244,7 +244,7 @@ void SummaryReader::findDeviations(std::vector<double>& absdev_vec,std::vector<d
       }
       else if((*referance_vec)[ivar]<(*checking_vec)[jvar]){
 	// Check with Linear polized arguments, jvar
-	double time_array[3];
+	double time_array[3]; // should contain { [time of occurance after referance] , [time of occurance before referance] , [time of referance] }
```
akva2: s/referance/reference, s/occurance/occurrence

tests/test_compareSummary.cpp
```diff
@@ -0,0 +1,69 @@
...
++   OPM is distributed in the hope that it will be useful,
++   but WITHOUT ANY WARRANTY; without even the implied warranty of
++   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
++   GNU General Public License for more details.
++
++   You should have received a copy of the GNU General Public License
++   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
++   */
+#include <opm/output/eclipse/summaryReader.hpp>
+#include "config.h"
```
akva2, line 20: config.h should always go first.

opm/output/eclipse/summaryReader.cpp
kristfho: I do not understand this comment.

akva2: sorry. it means replace referance with reference, and occurrance with occurrence. i.e. wrong spellings.

kristfho: Oh, thanks!

kristfho:
Hi, I'm having some problems with the function SummaryReader::close(). When I run the application i get "Aborted (core dumped)".

When running the application without using `close()` i have no trouble. Do the `ecl_sum_type` and `stringlist_type*` "clean up its own mess"?

opm/output/eclipse/summaryReader.cpp
atgeirr: Side note: the s/a/b notation comes from `sed` and similar tools using regular expressions. I think we cannot assume everyone is a grognard like you and me @akva2 :-)

akva2:
you should clean up the mess.

1) build debug
2) run binary through 'gdb'
3) see where it crashes

also useful: run debug binary through 'valgrind'

opm/output/eclipse/summaryReader.hpp
```diff
@@ -0,0 +1,80 @@
...
+public:
+  SummaryReader(){} 
+
+  SummaryReader(double relative_tolerance_max, double relative_tolerance_median_max):
+    relative_tolerance_max(relative_tolerance_max),
+    relative_tolerance_median_max(relative_tolerance_median_max)
+  {} 
+
+
+  bool open(const char* smspecFile1, const stringlist_type* unsmryFile1, const char* smspecFile2, const stringlist_type* unsmryFile2);
```
atgeirr: It would be much easier to use this class it all ert details could be encapsulated inside this class. At this point, we are almost there, except for the use of stringlist_type. Also, the SMSPEC and UNSMRY file names are the same except for the case name, so perhaps it is sufficient to just pass the two case names (as const std::string&) to this method? Also, open() and close() should essentially be (or be called from) the constructor and destructor, and not require the user to manually deal with this.

opm/output/eclipse/summaryReader.cpp
```diff
@@ -0,0 +1,333 @@
...
+  }
+  else {
+    return true;
+  }
+}
+
+void SummaryReader::close() {
+  ecl_sum_free_data(ecl_sum1);
+  ecl_sum_free_data(ecl_sum2);
+  stringlist_free(keys1);
```
akva2: these are not guaranteed to be allocated. stringlist_free does not check for a nullptr. if()'s required.

akva2: ignore this. i didn't notice the initializer in the header.

opm/output/eclipse/summaryReader.hpp
akva2: yes, both of these were issues i was just about to raise.

the signature of the constructor should be something along the lines of

``` C++
SummaryReader(double, double, const std::string& summaryFile1, const std::string& summary_file2);
```

in fact, there is no need to expose the ert usage in this class to the outside world at all.
since the ert types are all in form of pointers, we can delay their definition using prototyping. see https://github.com/akva2/opm-output/tree/compare_summary_ert_encapsulate

this is great, because should you for some reason want to drop ert, you don't have to change the user code.

akva2:
some suggested cleanups in https://github.com/akva2/opm-output/tree/compare_summary_warnings_cleanup

opm/output/eclipse/summaryReader.cpp
bska: > the `s/a/b` notation comes from sed and similar tools using regular expressions

Actually, it's older still.  The `s///` syntax for pattern matching and replacement originated in [ed](https://www.gnu.org/fun/jokes/ed-msg.html)--from which `sed` was created to be able to handle  files larger than could reasonably fit into memory and to automate common editing tasks.

jokva: ed is the **standard**

opm/output/eclipse/summaryReader.cpp
```diff
@@ -0,0 +1,333 @@
...
+#include <opm/output/eclipse/summaryReader.hpp>
+
+
+
+
+bool SummaryReader::open(const char* smspecFile1, const stringlist_type* unsmryFile1, const char * smspecFile2, const stringlist_type* unsmryFile2){
+  //opens the files
+  
+  
+  ecl_sum1 = ecl_sum_fread_alloc(smspecFile1, unsmryFile1, ":");
```
joakim-hove: If you use the function `ecl_sum_fread_alloc_case( )` you only need to provide one file - with optional extension, and the function will inspect the filesystem to locate the correct `SMSPEC / UNSMRY / .Snnnn` files.

opm/output/eclipse/summaryReader.cpp
```diff
@@ -0,0 +1,333 @@
...
+  if (ecl_sum1 == nullptr || ecl_sum2 == nullptr) {
+    OPM_THROW(std::runtime_error, "Not able to open files");
+  }
+  else {
+    return true;
+  }
+}
+
+void SummaryReader::close() {
+  ecl_sum_free_data(ecl_sum1);
```
joakim-hove: Here you should use `ecl_sum_free( )`.

opm/output/eclipse/summaryReader.hpp
kristfho: Thanks, this was helpful.

opm/output/eclipse/summaryReader.cpp
kristfho: Thanks, works perfectly!

kristfho:
I'm working on a integration test, thus it was practical to separate the code into a super class, and (soon two) sub classes.

kristfho:
In the latest commits most of the functions in the classes have been moved from public->private/protected.

I consider the code ready for a review, @akva2 .

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+
+#include <opm/output/eclipse/summaryComparator.hpp>
+#include <ert/ecl/ecl_sum.h>
+#include <ert/util/stringlist.h>
+#include <ert/util/int_vector.h>
+#include <ert/util/bool_vector.h>
+#include <opm/common/ErrorMacros.hpp>
+
+bool SummaryComparator::open(const char* basename1, const char* basename2){
```
akva2: i think this should be the constructor. it's a throw or true so there's no reason for it to not be, is there?

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+  ecl_sum2 = ecl_sum_fread_alloc_case(basename2, ":");
+  if (ecl_sum1 == nullptr || ecl_sum2 == nullptr) {
+    OPM_THROW(std::runtime_error, "Not able to open files");
+  }
+  else {
+    return true;
+  }
+}
+
+void SummaryComparator::close() {
```
akva2: destructor ?

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+
+
+
+//Read the data from the two files into separate vectors. Not necessary the same amount of values, but the values correspond to the same interval in time. Thus possible to interpolate values.
+ 
+void SummaryComparator::getDataVecs(std::vector<double> &dataVec1,std::vector<double> &dataVec2,const char* keyword){  
+  for (int time_index = 0; time_index < ecl_sum_get_data_length(ecl_sum1); time_index++){
+    dataVec1.push_back(ecl_sum_iget(ecl_sum1, time_index, ecl_sum_get_general_var_params_index( ecl_sum1 , keyword )));
```
akva2: since this vector potentially can have some size i think you should use

``` C++
dataVec1.reserve(ecl_sum_get_data_length(ecl_sum1);
```

to avoid repeated memory reallocation during the `push_back()` calls.

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+
+double SummaryComparator::average(std::vector<double> &vec){
+  return std::accumulate(vec.begin(), vec.end(), 0.0) / vec.size();
+}
+
+
+
+double SummaryComparator::linearPolation(double check_value, double check_value_prev , double time_array[3]){
+  //does a Linear Polation
```
akva2: interpolation?

opm/output/eclipse/summaryComparator.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+
+
+struct Deviation {
+  double abs; 
+  double rel; 
+};
+
+
+class SummaryComparator {
```
akva2: this is a public interface. i expect doxy on at least the part exposed to the world.

akva2:
additional request: my OCD twitched. i think we should stick the library-files in an opm/test_util/ folder to keep it separate from the actual writing code.

opm/output/eclipse/summaryComparator.cpp
kristfho: There is a destructor in the declaration of the class (in the header file), which call the close()-function inline.

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+    dataVec1.push_back(ecl_sum_iget(ecl_sum1, time_index, ecl_sum_get_general_var_params_index( ecl_sum1 , keyword )));
+  }
+  for (int time_index = 0; time_index < ecl_sum_get_data_length(ecl_sum2); time_index++){
+    dataVec2.push_back(ecl_sum_iget(ecl_sum2, time_index, ecl_sum_get_general_var_params_index( ecl_sum2 , keyword )));
+  }
+   
+}
+
+
+void SummaryComparator::setDataSets(std::vector<double> &timeVec1,std::vector<double> &timeVec2){
```
joakim-hove: I would recommend calling the two cases e.g. `test_case` and `ref_case` and then _stick to that_. The number of timesteps in a simulation is an "implementation detail" - and might change during development, without that really being an error of any kind - I then think you risk that the roles of `case1` and `case2` with respect to `long` and `short` might change - that might very well cause confusion.

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+    ecl_sum_fileShort = this->ecl_sum2;
+    ecl_sum_fileLong = this->ecl_sum1;
+  }
+}
+   
+
+
+//Figures out which time vector that contains the fewer elements. Sets this as referenceVec and its corresponding 
+// data as referenceDataVec. The other vector-set as checkVec( the time vector) and checkDataVec.
+void SummaryComparator::chooseReference(std::vector<double> &timeVec1,std::vector<double> &timeVec2,std::vector<double> &dataVec1,std::vector<double> &dataVec2){
```
joakim-hove: See: https://github.com/OPM/opm-output/pull/56/files#r70948834

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+//Calculates each deviation and stores it in a deviation struct. Uses the referenceVec as basis, and checks its values against the values int checkDataVec
+//Should be called inside a for/while loop running over the reference vector, with ivar as the iterative variable of the loop. 
+// Updates iteration variable jvar aswell, which will be used for the checkVec/cecheDataVec, thus will not check same keyword twice
+void SummaryComparator::getDeviation(int refIndex, int &checkIndex, Deviation &dev){
+  if((*referenceVec)[refIndex] == (*checkVec)[checkIndex]){
+    dev = SummaryComparator::calculateDeviations((*referenceDataVec)[refIndex], (*checkDataVec)[checkIndex]); 
+    checkIndex++;
+    return;
+  }
+  else if((*referenceVec)[refIndex]<(*checkVec)[checkIndex]){
```
joakim-hove: This I do not understand.

opm/output/eclipse/summaryComparator.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+
+
+double SummaryComparator::average(std::vector<double> &vec){
+  return std::accumulate(vec.begin(), vec.end(), 0.0) / vec.size();
+}
+
+
+
+double SummaryComparator::linearPolation(double check_value, double check_value_prev , double time_array[3]){
```
joakim-hove: There does not seem to be any calls to this method - so it might not be a problem at all. But - interpolation of summary variables (which I kind of assumed this was all about?) is slightly non trivial. You must take the _type_ of the variable into account and either do straight linear interpolation, or "constant interpolation" - see a discussion of the topic [here](https://github.com/Ensembles/ert/blob/master/devel/python/python/ert/ecl/ecl_sum.py#L1076) and [here](https://github.com/Ensembles/ert/blob/master/devel/libecl/src/ecl_sum_data.c#L146).

The ert function `ecl_sum_get_interp_sim_time()` does exactly this.

opm/output/eclipse/summaryComparator.cpp
kristfho: I don't fully understand what you are trying to say here. In the current constructor. there is a call for `init(const, char*, const char*)`  which calls this open(..)-function. Do you want the functionallity of the open-function implemented directly in the constructor (and no open-function)?

opm/output/eclipse/summaryComparator.cpp
akva2: precisely. unless you think there are situations where it makes sense to have construction and reading detached i strongly prefer to construct in the constructor.

opm/output/eclipse/summaryComparator.cpp
akva2: precisely. unless you think there are situations where it makes sense to have construction and reading detached i strongly prefer to construct in the constructor.

opm/output/eclipse/summaryRegressionTest.cpp
```diff
@@ -0,0 +1,125 @@
...
+void RegressionTest::getRegressionTest(const char* keyword){
+  std::vector<double> timeVec1, timeVec2;
+  setTimeVecs(timeVec1, timeVec2);  // Sets the time vectors, they are equal for all keywords (WPOR:PROD01 etc)
+  setDataSets(timeVec1, timeVec2); //Figures which dataset that contains more/less values pr keyword vector.
+  
+  if(stringlist_contains(keysShort,keyword) && stringlist_contains(keysLong, keyword)){
+    checkForKeyword(timeVec1, timeVec2, keyword);
+    return;
+  }
+  std::cout << "The keyword suggested, " << keyword << ", is not supported by one or both of the summary files. Please use a different keyword." << std::cout;
```
akva2: last cout should be endl.

opm/output/eclipse/summaryComparator.cpp
kristfho: I put the close-function's functionality in the destructor and removed the close-function.

opm/output/eclipse/summaryComparator.cpp
kristfho: It says that if both the check file and reference file have a data value for a specific time `(*referenceVec)[refIndex] == (*checkVec)[checkIndex])` compare these data. 

If the check file's time-value does not match the reference files time-value, we have to guess the value of one of the files to be able to say anything about the deviation. We might use interpolation, however at this moment the way of guessing is a unit step function, we are basically saying that when the value is unknown, we use the latest known value. 

It is important to have in mind that the proper use of the getDeviation-function, requires a loop which iterates over the reference file, the file with the   fewest number of time steps. It could have been done opposite, that the reference file have the most time steps, however this will make the "resolution" of the check file worse. What i mean is that when the time-values do not match, there best option is to use the file with the most time steps to guess the right value for the time step, rather than to guess the value in the file containing fewer time steps.

opm/output/eclipse/summaryComparator.cpp
kristfho: So far the time vectors have been of constant size, meaning that all the keywords of a file would have the same number of time steps (I have tested mostly on Norne and SPE9_CP). Is it wrong to assume that this holds in general?

opm/output/eclipse/summaryComparator.cpp
joakim-hove: OK - I understand; I would _really_ recommend doing the comparison based on interpolation on a controlled regular time axis as suggested [here](https://github.com/OPM/opm-common/pull/154#issuecomment-229014482) (ignore the Python) - your call.

opm/output/eclipse/summaryComparator.cpp
joakim-hove: > Is it wrong to assume that this holds in general?

No - that is correct; for a given summary output all the keywords e.g. `WWCT:OP_1` and `FOPR` share the same time axis. However - the number of timesteps in a simulation is determined by the simuator, and will typically vary with the keywords used - and in the case of `flow` also with the implementation.

opm/output/eclipse/summaryComparator.cpp
kristfho: Well, the the idea of using the unit step approach was @alfbr 's. 
The reason for using the unit step is that: As far as I have understood, when the well experiences a major change in something (not sure exactly what), the simulators writes out the time step. Say the major change is that the well starts producing, change will be from 0 to something. The relative change from zero to something is 1. If it is in this area the mismatch in reporting time occurs, the use of interpolation will trigger the deviation test. The test will fail based on a false error. By using a unit step approach, this will not happen.

However, in other time intervals, without such critical changes, the interpolation will "make a better guess" than the unit step. Maybe I should include logic that use both? Say I first use interpolation, and if that one fails, use unit step and check if it gives an acceptable deviation?

joakim-hove: > Well, the the idea of using the unit step approach was @alfbr 's. 

Does not necessarily imply it is a freaking good idea:-) Anyway - I'll leave you to it.

opm/output/eclipse/summaryComparator.cpp
kristfho: If i understand you correctly, all summary vectors corresponding to different keywords e.g. `WWCT:OP_1` `WOPR:PRODU1` etc will have the same amount of data values and time steps. File1 might have 91 values and 91 timesteps pr keyword, while File2 has  93 values and timesteps. 

If this is the case, do I not understand how this is a problem.

> I then think you risk that the roles of case1 and case2 with respect to long and short might change

joakim-hove: > File1 might have 91 values and 91 timesteps pr keyword, while File2 has 93 values and timesteps.

Yes; let us further assume that `File1` correponds to the simulation which corresponds to case we are interested in testing i.e. `flow` and that `File2` corresponds to the reference solution. Then `File1` will be the `shortCase` and `File2` the `longCase` - fine.

I foresee that at some later stage `flow` changes and now suddenly `File1` has 95 timesteps, then the role `longCase` and `shortCase` will be interchanged - if the code is freaking perfect that should pose no problems, but I just guess that this sudden change - which might take place three months down the line, might create trouble.

kristfho:
Note:  
- Do not know if the doxy comments are correct. (just followed a tutorial i found online).
- Have not included both interpolation and unit step in the logic

opm/output/eclipse/summaryComparator.hpp
```diff
@@ -33,59 +34,58 @@ typedef struct ecl_sum_struct ecl_sum_type;
...
+  std::vector<double> * referenceDataVec = nullptr;///< For storing the data values corresponding to each time step for the file containing the fewer time steps
+  std::vector<double> * checkVec = nullptr;///< For storing the values of each time step for the file containing the more time steps
+  std::vector<double> * checkDataVec = nullptr;///< For storing the data values corresponding to each time step for the file containing the more time steps
+  
+  void getDeviation(int refIndex, int &checkIndex, Deviation &dev); /**<Calculates each deviation and stores it in a Deviation struct. Uses the referenceVec as basis, and checks its values against the values int checkDataVec. 
+								     *Should be called inside a for/while loop running over the reference vector, with ivar as the iterative variable of the loop.
+								     *Updates iteration variable jvar aswell, which will be used for the checkVec/cecheDataVec, thus will not check same keyword twice*/
+ 
+  void setDataSets(std::vector<double> &time_vec1,std::vector<double> &time_vec2);///< Figures out which data file that contains the more/fewer time steps, and gives it the correct pointer.
```
akva2: better. but one example how i think the doxy should look:

```
//! \brief Figure out which data file contains the most / less timesteps and assign internal pointers accordingly.
//! \param[in] time_vec1 Data from first file
//! \param[in] time_vec2 Data from second file
```

opm/output/eclipse/summaryComparator.hpp
```diff
@@ -33,59 +34,58 @@ typedef struct ecl_sum_struct ecl_sum_type;
...
-  void getDataVecs(std::vector<double> &dataVec1,std::vector<double> &dataVec2, const char* keyword); 
-void chooseReference(std::vector<double> &timeVec1,std::vector<double> &timeVec2,std::vector<double> &dataVec1,std::vector<double> &dataVec2);
-  double getRelTolerance(){return this->relativeTolerance;}
-  double getAbsTolerance(){return this->absoluteTolerance;}
+  std::vector<double> * referenceVec = nullptr; ///< For storing the values of each time step for the file containing the fewer time steps
+  std::vector<double> * referenceDataVec = nullptr;///< For storing the data values corresponding to each time step for the file containing the fewer time steps
+  std::vector<double> * checkVec = nullptr;///< For storing the values of each time step for the file containing the more time steps
+  std::vector<double> * checkDataVec = nullptr;///< For storing the data values corresponding to each time step for the file containing the more time steps
+  
+  void getDeviation(int refIndex, int &checkIndex, Deviation &dev); /**<Calculates each deviation and stores it in a Deviation struct. Uses the referenceVec as basis, and checks its values against the values int checkDataVec.
```
akva2: another example;

```
//! \brief Calculate deviation between two data values and stores it in a Deviation struct.
//! \param[in] refIndex Index in reference data
//! \param[in] checkindex Index in data to be checked.
//! \param[out] dev Holds the result from the comparison on return.
//! \details Uses the referenceVec as basis, and checks its values against the values in checkDataVec.
```

akva2: note: execute 'make doc' to build the documentation. it will generate in a 'doc' subfolder in the build directory, html output is the most interesting one.

opm/output/eclipse/summaryComparator.cpp
alfbr: Of course not  :-) I still suggest we go with a simple step function first and see how that works out.  We may not need more sophistication. Since pressure and rate curves has discontinuities at well changes, a simple linear interpolation will not work.

opm/output/eclipse/summaryComparator.hpp
kristfho: For the member variables, which syntax do you recommend? 
At the moment it is implemented like this: 

``` C++
stringlist_type * keys1 = nullptr; ///< For storing the values of each time step for the file containing the fewer time steps.
```

akva2: that looks good to me. if i am to be picky i tend to prefer //!<,

``` C++
stringlist_type * keys1 = nullptr; //!< Keys in file containing fewer time steps.
```

akva2:
contained in #55

PR #61, opened 2016-07-07
Overwrite UNRST file even when 0 isn't first step

jokva:
https://github.com/OPM/opm-simulators/issues/753
Under some conditions, such as the case reported in issue-753, UNRST
files would be appended to rather than opened for writing (which would
overwrite them).

The Restart class is now the representation of restart files as a
concept, whereas Restart::File represents the phyiscal file with the add
operations. File has adopted Restart's writing interface.


jokva:
I'd like to add that this still isn't a complete fix. It doesn't handle the case where you restart from a simulation, then keep on simulating - it would simply erase the first N report steps. This should be addressed before a merge.

pgdr:
This looks very good!  Do you propose that someone constructs a second PR that builds on this to solve the restart-and-keep-simulation case that you mention, or are you working on that now?

jokva:
I haven't worked on it at all, actually, because I've had to mature the idea a bit and I actually have now idea how we want it to behave. It needs some discussion (really waiting for @joakim-hove and others). That, and since restarting flow seems broken anyway I figured there was no rush.

cfbe:
I seem to have a similar problem as reported in OPM/opm-simulators#753.

Will this pull-request fix this issue?

When I run for the first time:

```
tr-lx627802:~/cc/opm/opm-data/spe3/test:$ ll
total 66
-rw-rw---- 1 cfbe cfbe 15234 May 26 08:43 SPE3CASE1.DATA
tr-lx627802:~/cc/opm/opm-data/spe3/test:$ flow SPE3CASE1.DATA 
```

I can visualize in Resinsight without a problem.

```
tr-lx627802:~/cc/opm/opm-data/spe3/test:$ ll *.UNRST
-rw-rw---- 1 cfbe cfbe 134176 Sep  8 10:57 SPE3CASE1.UNRST
```

However, when I run flow for the second time, it has double up the size and will not visualize:

```
tr-lx627802:~/cc/opm/opm-data/spe3/test:$ ll *.UNRST
-rw-rw---- 1 cfbe cfbe 268352 Sep  8 10:58 SPE3CASE1.UNRST
```

jokva:
It will fix it in the sense that it changes behaviour to always overwrite if you're using UNIFOUT. However, that means if you're trying to start the simulation from a restart file it will **overwrite** rather than append (as it probably should in that case). If you're only doing fresh simulations instead of attempting to restart then yes, it will solve the problem.

jokva:
A (maybe good) workaround could be to ask for the files in non-unified format, which would overwrite the steps in question, and then post-process with the ert `ecl_pack.x` tool to produce your unified file.

cfbe:
A workaround like the one you suggest seems complicated. To re-run a simulation is something that is conducted routinely. For that reason I hope one could avoid any workaround.

jokva:
It's not any more complicated than `flow CASE.DATA && ecl_pack.x CASE`. Alias it to `flow-pack` and you're good to go.

When you say re-run, do you mean without attempting to start from a previously run simulation? Without attempting to restart, this patch (once cleaned up) would solve your problem, as it will always override your previous output.

jokva:
Fixing it _is_ on the agenda, though, it's just not that much of a priority yet, unfortunately.

cfbe:
Yes, by rerun I mean running without a restart, starting all over from the start. A typical work process:
flow CASE.DATA
check results
edit CASE.DATA
flow CASE.DATA
It is good to hear that this fix is on the agenda. I believe this is an important issue that needs prioritization.

jokva:
@cfbe Can you try with https://github.com/OPM/opm-output/pull/89 and check if your issue still persists?

cfbe:
Hi,
I have now tried #89 , and it works fine now.

PR #63, opened 2016-07-12
WIP: Fwrite init

joakim-hove:
Will write the DEPTH and DX, DY, DZ keywords to the INIT file.

Requires:  Ensembles/ert#1164

### DIFF
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -46,6 +46,7 @@
 #include <ert/ecl/ecl_kw_magic.h>
 #include <ert/ecl/ecl_init_file.h>
 #include <ert/ecl/ecl_file.h>
+#include <ert/ecl/ecl_grid.h>
 #include <ert/ecl/ecl_rft_file.h>
 #include <ert/ecl/ecl_rst_file.h>
 #include <ert/ecl_well/well_const.h>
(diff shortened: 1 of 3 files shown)


jokva:
Why is the `fwrite` feature on the grid itself? Isn't the grid largely data representation only?

joakim-hove:
> Why is the fwrite feature on the grid itself? Isn't the grid largely data representation only?

Well; I guess it is _mainly_ data representation - but the data we are talking about here is not stored anywhere - it is calculated on the demand in the `ecl_grid_fwrite_DIMS()` and `ecl_grid_fwrite_DEPTH()` functions.

jokva:
Sure, but that's just an implementation detail and doesn't affect (or shouldn't rather) the interface.

These functions don't really offer anything over just the raw ERT call, and they don't really rely on the grid internals. I think the entire implementation should go in output.

joakim-hove:
>  I think the entire implementation should go in output.

Meaning that you think: https://github.com/OPM/opm-parser/pull/874 should be removed? That is certainly possible - but then we need to expose the underlying `ecl_grid_type`pointer from the `EclipseGrid` class all the way out to output; alternatively _everything_ could be in opm-output - including the functionality covered by: https://github.com/Ensembles/ert/pull/1164 - that would certainly work.

jokva:
Pretty much. As far as I can tell the PR in parser effectively does nothing.

alfbr:
Not sure what _everything_ means here, but if feasible, it would be very nice to have INIT writing in opm-output.

jokva:
I'm ok with ert providing the actual writing and grid computation (which is rather ert internal anyway).

joakim-hove:
> I'm ok with ert providing the actual writing and grid computation (which is rather ert internal anyway).

Well - that was exactly my point as well. 

> Not sure what everything means here, but if feasible, it would be very nice to have INIT writing in opm-output.

Well - of course it will be invoked and operated from opm-output; the question is whether the implementation should be 100% in opm-output or not. In my opinion it was quite a lot cleaner to let opm-output call out to ert to do the lowest level of writing/computation - but I agree with @jokva that the Parser PR: OPM/opm-parser#874 is unneccessary.

Lunch now - fixup coming.

jokva:
Nice. Feel free to squash too.

joakim-hove:
OK - I'll merge this when/if Travis is happy.

PR #65, opened 2016-07-15
Completion production/injection rates/totals; Field keywords

jokva:
A small set of the completion family of keywords, water/oil/gas
production and injection. The tests and example data file are updated
accordingly, with edge cases.

Requires https://github.com/Ensembles/ert/pull/1167 and https://github.com/OPM/opm-parser/pull/881 (bugfixes). 

note: I've added support for a set of Field keywords to this PR. Easily split up obviously, but they're along the same lines so I figured  it's fine.

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -121,21 +121,22 @@ inline quantity rate( const fn_args& args ) {
 
 template< rt phase, bool injection = true >
 inline quantity crate( const fn_args& args ) {
-    double sum = 0.0;
+    const quantity zero = { 0, rate_unit< phase >() };
+    const auto index = args.active_index;
 
-    for( const auto* sched_well : args.schedule_wells ) {
-        const auto& name = sched_well->name();
-        if( args.wells.wells.count( name ) == 0 ) continue;
-        const auto& well = args.wells.at( name );
-        if( well.completions.count( args.active_index ) == 0 ) continue;
-        const auto v = well.completions.at( args.active_index ).rates.get( phase, 0.0 );
-        if( ( v > 0 ) == injection )
-            sum += v;
-    }
+    if( args.schedule_wells.empty() ) return zero;
 
-    if( !injection ) sum *= -1;
+    const auto& name = args.schedule_wells.front()->name();
+    if( args.wells.wells.count( name ) == 0 ) return zero;
 
-    return { sum, rate_unit< phase >() };
+    const auto& well = args.wells.at( name );
+    if( well.completions.count( index ) == 0 ) return zero;
+
+    const auto v = well.completions.at( index ).rates.get( phase, 0.0 );
+    if( ( v > 0 ) != injection ) return zero;
+
+    if( !injection ) return { -v, rate_unit< phase >() };
+    return { v, rate_unit< phase >() };
 }
 
 template< rt phase > inline quantity prodrate( const fn_args& args )
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -77,13 +77,14 @@ struct quantity {
  * All functions must have the same parameters, so they're gathered in a struct
  * and functions use whatever information they care about.
  *
- * ecl_wells are wells from the deck, provided by opm-parser. wells is
- * simulation data
+ * ecl_wells are wells from the deck, provided by opm-parser. active_index is
+ * the index of the block in question. wells is simulation data.
  */
 struct fn_args {
     const std::vector< const Well* >& ecl_wells;
     double duration;
     size_t timestep;
+    int active_index;
     const data::Wells& wells;
 };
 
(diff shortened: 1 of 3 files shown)


opm/output/eclipse/Summary.cpp
```diff
@@ -77,13 +77,14 @@ struct quantity {
  * All functions must have the same parameters, so they're gathered in a struct
  * and functions use whatever information they care about.
  *
- * ecl_wells are wells from the deck, provided by opm-parser. wells is
- * simulation data
+ * ecl_wells are wells from the deck, provided by opm-parser. active_index is
+ * the index of the block in question. wells is simulation data.
  */
 struct fn_args {
     const std::vector< const Well* >& ecl_wells;
```
joakim-hove: As in [this well](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Schedule/Well.hpp) - maybe `schedule_wells` would be a better name?

opm/output/eclipse/Summary.cpp
```diff
@@ -102,18 +103,51 @@ measure rate_unit< rt::gas >() { return measure::gas_surface_rate; }
...
 inline quantity rate( const fn_args& args ) {
     double sum = 0.0;
 
     for( const auto* ecl_well : args.ecl_wells ) {
-        if( args.wells.wells.count( ecl_well->name() ) == 0 ) continue;
-        sum += args.wells.at( ecl_well->name() ).rates.get( phase, 0.0 );
+        const auto& name = ecl_well->name();
+        if( args.wells.wells.count( name ) == 0 ) continue;
+        const auto v = args.wells.at( name ).rates.get( phase, 0.0 );
+        if( ( v > 0 ) == injection )
```
joakim-hove: I would spend less mental energy on ` if ( ( v > 0) && injection)`.

joakim-hove, line 114: There are four states here `{ ( v >  0) , ( v < 0 )} x { true , false }` - but I assume only `{(v > 0) , true }` and `{ (v < 0) , false }` are valid - otherwise you have a massive internal error? So I guess you just do `v += sum`?

opm/output/eclipse/Summary.cpp
```diff
@@ -102,18 +103,51 @@ measure rate_unit< rt::gas >() { return measure::gas_surface_rate; }
...
+    }
+
+    if( !injection ) sum *= -1;
+
+    return { sum, rate_unit< phase >() };
+}
+
+template< rt phase > inline quantity prodrate( const fn_args& args )
+{ return rate< phase, false >( args ); }
+template< rt phase > inline quantity injerate( const fn_args& args )
```
joakim-hove, line 144: A blank line maybe?

opm/output/eclipse/Summary.cpp
```diff
@@ -102,18 +103,51 @@ measure rate_unit< rt::gas >() { return measure::gas_surface_rate; }
...
+        if( ( v > 0 ) == injection )
+            sum += v;
     }
 
+    if( !injection ) sum *= -1;
     return { sum, rate_unit< phase >() };
 }
 
+template< rt phase, bool injection = true >
+inline quantity crate( const fn_args& args ) {
```
joakim-hove, line 123: For the hierarchy `well -> group -> field` I guess it makes sense to express everything as sums over well-lists, but when it comes to the connections/completions I do not see the point. The `Cxxx` variables are always only interesting as variables for one completion - and are never used as elements in a sum constituting a "larger object" - the way wells are summed into groups. Does this make sense?

joakim-hove:
I will force merge this: OPM/opm-simulators#763 tomorrow and restart Travis.

opm/output/eclipse/Summary.cpp
jokva: Good one, I'll update in the morning.

opm/output/eclipse/Summary.cpp
jokva, line 114: Pretty much. These are the cases:
1. Injection is requested and this is an injecting well (rates are positive).
2. Injection is requested and this is a producer.
3. Production is requested and this is a producer.
4. Production is requested and this is an injector.

Which gives us:
1. `v > 0 == true && injection == true` -> sum increases
2. `v > 0 == true && injection != true )` -> no-op
3. `v > 0 == false && injection != true` -> sum increases
4. `v > 0 == false && injection == true` no-op

So it's a no-op when the well type doesn't matche the boolean parameter.

opm/output/eclipse/Summary.cpp
jokva, line 144: Maybe. I really wish C++ had reasonable function aliasing. Figured it wasn't worth spending too much space on.

opm/output/eclipse/Summary.cpp
jokva, line 123: Yeah. I mean, we can throw in an assert I suppose, but the big reason for just reusing everything was that wells already come in a list, so it would be zero or one iteration. Could probably use less noise though, I agree with that, so I'll look into it.

jokva:
Excellent. Works on my machine, but you never know. I'll address the issues tomorrow.

opm/output/eclipse/Summary.cpp
jokva, line 123: I've updated this. It's actually slightly longer (I **really** wish we had monads), but it does communicate the no-sum property of the completion keywords.

PR #66, opened 2016-07-19
Write units to Summary file.

jokva:
Support for writing units in the UNITS part of the SMSPEC header.

This triggered a few small refactorings: The quantity arithmetics deals with the necessary unit conversions, largely simplifying the DSL because types are "inferred" rather than explicitly converted. Restores commutative and associative properties of multiplication (where multiplication is used and/or needed, not in general).

Support for the FWCTH keyword is added. This should've already been implemented, but was an oversight on my part.

Requires https://github.com/OPM/opm-parser/pull/885


jokva:
I'll restart travis here.

PR #68, opened 2016-07-21
Revert "C++ regression test of eclipse files (initial, unified restart and summary)"

jokva:
Reverts OPM/opm-output#55


akva2:
grow up.

PR #70, opened 2016-07-26
Integration tests and improved regression tests for Eclipse files

ptaule:
In this PR, the comparison library files EclFilesComparator.cpp and EclFilesComparator.hpp support comparison of RFT files and also supports comparison of more keyword types (bool, int and char keyword types). Also there are some minor changes to the summary compare libraries and the applications in test_util/.


opm/test_util/EclFilesComparator.cpp
```diff
@@ -35,105 +35,147 @@ bool ECLFilesComparator::keywordValidForComparing(const std::string& keyword) co
...
-void ECLFilesComparator::deviationsForOccurence(const std::string& keyword, int occurence) {
+void ECLFilesComparator::intComparisonForOccurence(const std::string& keyword, int occurence) const {
     ecl_kw_type* ecl_kw1         = ecl_file_iget_named_kw(ecl_file1, keyword.c_str(), occurence);
     ecl_kw_type* ecl_kw2         = ecl_file_iget_named_kw(ecl_file2, keyword.c_str(), occurence);
     const unsigned int numCells1 = ecl_kw_get_size(ecl_kw1);
     const unsigned int numCells2 = ecl_kw_get_size(ecl_kw2);
     if (numCells1 != numCells2) {
         std::cout << "For keyword " << keyword << " and occurence " << occurence << ":\n";
-        OPM_THROW(std::runtime_error, "Number of active cells are different.");
+        OPM_THROW(std::runtime_error, "Number of cells are different.");
```
joakim-hove: Here copying gives you nothing compared to `ecl_kw_equal (..)`.

joakim-hove: The same applies to the string and bool based comparisons.

joakim-hove:
A .RFT file can contain three different types of RFTs:
1. Plain RFT.
2. PLT
3. PLT with segment information.

Opm currently only supports 1).   The ert ecl_rft_file abstracts away the difference between 1 and 2 - to the enten possible. So using that it will be easier to compare Opm and Eclipse results in the (quite common) case that Eclipse has written a PLT.

ptaule:
Ok, thanks for the info! This is a regression test i.e. flow vs flow, so I guess it might be beneficial to be able to compare all cases if flow should support 2. and 3. in the future(?) I will, however, look into it for the integration test. 

> -- to the enten possible.

What do you mean by this?

joakim-hove:
To the _extent possible_ it should be.

ptaule:
This is ready for review.

joakim-hove:
It is not clear to me who is supposed to review and merge this - but if it is me I will look into it on monday.

Right now the Travis build is red, it is not clear to me why that is the case,  but the PR will not be merged before Travis is green - or dependencies to other PR's have been clearly explained.

ptaule:
I think the reason why the Travis build fails is that this patch renames the applications, which makes the ctest added here OPM/opm-simulators#754 fail. It should be fixed with OPM/opm-simulators#771 and OPM/opm-common#161 though.

joakim-hove:
> It is not clear to me who is supposed to review and merge this

By private communication from @alfbr this will be handled by @akva2.

opm/test_util/EclFilesComparator.cpp
```diff
@@ -305,3 +217,402 @@ double ECLFilesComparator::average(const std::vector<double>& vec) {
...
+            if (std::abs(val2) > absTolerance) {
+                printValuesForCell(keyword, occurrence1, occurrence2, cell, val1, val2);
+                OPM_THROW(std::runtime_error, "Negative value in second file, "
+                        << "which in absolute value exceeds the absolute tolerance of " << absTolerance << ".");
+            }
+            val2 = 0;
+        }
+    }
+    Deviation dev = calculateDeviations(val1, val2);
+    if (dev.abs > absTolerance && dev.rel > relTolerance) {
```
joakim-hove, line 324: I guess this should be: `if (dev.abs > absTolerance || dev.rel > relTolerance)` - i.e. we should flag an error if _either_ the absolute or the relative deviation is outside the limits?

akva2, line 324: && was the intention. the idea is that the absolute test 'should' fail in general, so the relative governs the general behavior. the exception is around 0 where the relative test will fail in general, the absolute test then 'runs the show'.

opm/test_util/EclFilesComparator.cpp
```diff
@@ -305,3 +217,402 @@ double ECLFilesComparator::average(const std::vector<double>& vec) {
...
+    const double relDeviationAverage = average(relDeviation);
+    std::cout << "Average absolute deviation = " << absDeviationAverage  << std::endl;
+    std::cout << "Median absolute deviation  = " << median(absDeviation) << std::endl;
+    std::cout << "Average relative deviation = " << relDeviationAverage  << std::endl;
+    std::cout << "Median relative deviation  = " << median(relDeviation) << "\n\n";
+}
+
+
+
+void RegressionTest::boolComparisonForOccurrence(const std::string& keyword, int occurrence1, int occurrence2) const {
```
joakim-hove, line 237: The methods `RegressionTest::$typeComparisonForOccurence()` probably work just fine, but for the trivial `==` based comparisons a simpler alternatively can be implemented using the `ecl_kw_equal()` function. That way some code is shaved off - and the same implementation can be shared for `bool, char` and `int`.

akva2: i don't see how i can do the equivalent using ecl_kw_equal. it won't tell me where it differs.

joakim-hove: You can use `ecl_kw_first_different`, but I guess that for the not-floating point vectors it is either all  equal - or not.

But by all means, just a suggestion on my part.

akva2: a good idea and i appreciate it, but since it wouldn't be doing the same thing i'll keep the current code for now.

akva2:
replaced by #90

PR #71, opened 2016-07-27
c++ summaryComparator, [NOT FINISHED]

kristfho:
This PR contains a new sub class of SummaryComparator, IntegrationTest.
The binary that runs the regression test have been replaced by a binary that can run both regression tests and integration tests (via command line options).


opm/test_util/summaryIntegrationTest.cpp
```diff
@@ -0,0 +1,352 @@
...
+	    WellProductionVolume result =  getWellProductionVolume(keyword);
+	    WBHP.total += result.total; 
+	    WBHP.error += result.error; 
+	}	
+    }
+}
+
+
+void IntegrationTest::updateVolumeError(const char* keyword){
+    std::string keywordString(keyword);
```
joakim-hove, line 197: You do reinvent quite a lot of ert code - a bit unnecceassary in my opinion.
1. Based on the ert smspec_node_type you can query whether the variable represents a total (i. e  volume) and/or historical quantity.
2. You seem to find totals by integrating rates, I would really recommend using the already summed up xxxT _total_ keywords.
3. Integrating WBHP does not give a volume but rather something like momentum per area .

opm/test_util/summaryComparator.cpp
```diff
@@ -204,7 +204,72 @@ const char* SummaryComparator::getUnit(const char* keyword){
 }
 
 
+void SummaryComparator::printKeywords(){
+    size_t ivar = 0;
+    std::vector<std::string> noMatchString; 
+    std::cout << "Keywords that are common for the files:" << std::endl;
+	while(ivar < stringlist_get_size(keysShort)){
+	    const char* keyword = stringlist_iget(keysShort, ivar);           
+	    for (size_t jvar = 0; jvar < stringlist_get_size(keysLong); jvar++){
```
joakim-hove: The stringlist code has a function:

``` C
stringlist_contains ( ..., ...);
```

test_util/summaryComparator.cpp
```diff
@@ -0,0 +1,203 @@
...
+    
+    
+    int argOffset = optind;
+    if (argc != argOffset + 4) {
+        printHelp();
+        return EXIT_FAILURE;
+    }
+    const char * basename1   = argv[argOffset];
+    const char * basename2   = argv[argOffset+1];
+    double absoluteTolerance = atof(argv[argOffset+2]);
```
joakim-hove: `atof` is quite fragile when it comes to malformed input - maybe consider e.g. `strtod`

opm/test_util/summaryIntegrationTest.cpp
joakim-hove, line 197: Observe that iff you insist on determing total volumes by integrating rates the correct representation of the function R (t) is as a piecewise constant function,  where the constant value is given as the rate at the upper limit of the interval.

Again - you can query ert with the is_rate property to check the correct temporal interpolation behavior.

If the WEFAC/ GEFAC keywords are different from 1.0 the integration based solution will fail by construction (opm does not support these keywords and maybe never will - so this is maybe not so relevant).

opm/test_util/summaryComparator.cpp
kristfho: Thanks, changed to using the `stringlist_contains()` function.

test_util/summaryComparator.cpp
kristfho: oh, I didn't know. Changed to `strtod`.

opm/test_util/summaryIntegrationTest.cpp
kristfho, line 197: 1. I'm not sure how to use the smspec_node_type to get the information needed. 
2. I do agree that the best solution for finding e.g. the total production of a well is to use the last data value of the total keywords. This is what I do in the `volumeUpdateError(keyword)` function. The reason for integrating up the rate to a total is that it is the easiest way to check whether the error rate (total error/total production) is acceptable. I'm not saying it is the best way, however it is quite simple. An advantage is that it will not be possible to compare the total error from the well of interest with the total production from another well.  
3. > Integrating WBHP does not give a volume

True. I have commented on this in the `summaryIntegrationTest.hpp`. For simplicity, I will treat the integrated WBHP as a volume.

I'm considering removing the WellProductionVolume member variables of the class, and rather using double values, e.g. WOPT and WOPError.

joakim-hove, line 197:
``` C
smspec_node_type * node = ecl_sum_get_general_var_node (  );

bool total = smspec_node_is_total ( );
bool rate = smspec_node_is_rate (  );
bool hist = smspec_node_is_historical ( );

```

alfbr, line 197: > You do reinvent quite a lot of ert code - a bit unnecceassary in my opinion.

Not in this case I believe. ERT does not provide the integral of the WBHP curves to my knowledge, so this would need to be implemented somehow. When the same routine works for the rate curves, using calls to some code outside OPM seems only to obfuscate the code. Still, for such a simple operation, there should be no need to use an external routine which is limited to one file format regardless.

To appreciate the approach here, I think it is important to look at the intent. The intent of the implementation under discussion is to obtain some notion of when two simulations are close. Typically comparing simulations from two different simulators. In this case, for the summary part, we need to focus only on the main solution variables from the well equations. Hence, only production rates and BHP is checked. In this case we need not even have totals available. Checking that totals are calculated correctly is already catered to in the regression tests, here we want the test to work even when reported totals are wrong or not written out at all.

> Observe that iff you insist on determing total volumes by integrating rates the correct representation of the function R (t) is as a piecewise constant function,

Yes, so the integral should boil down to a simple Riemann sum.

> where the constant value is given as the rate at the upper limit of the interval.

This sounds wrong, but maybe I misunderstand here.

> Integrating WBHP does not give a volume but rather something like momentum per area .

Yes, so a renaming can be wise here. For instance removing word `volume` or replacing it by `integral` should make it clear. In either case, we do need to check relative errors, so the integral is necessary.

joakim-hove, line 197: 1) Well - my main point was the code related to determining whether this is a historical variable,  a rate variable or a total variable. That is currently implemented with some quite verbose string checks.

2 )Assume rates R1 and R2 at times T1 and T2 then the total production from T1 to T2 is given by R2 \* (T2 -T1) - _not_ (R1 + R2) \* (T2 -T1)/2 - i. e. the rate is only evaluated at the upper limit.

alfbr, line 197: > Assume rates R1 and R2 at times T1 and T2 then the total production from T1 to T2 is given by R2 \* (T2 -T1)

Thanks for the clarification, that is exactly what we should do.

kristfho:
Thanks, I will look into it. 

If i understand you correctly, the correct way to find total volumes is by summing rectangles, and for summing the pressure one should use a trapezoid method? 

> When the data corresponds to a rate, it is importaint to use the data value ...

Will the long comment be correct now?

joakim-hove:
Yes - that is in my opinion correct.

In general the trapezoidal rule, but for rates - which is a large and important exception, the rate should be assumed to be piecewise constant at the upper limit - i.e. the rectangle as you say.

alfbr:
> In general the trapezoidal rule, but for rates - which is a large and important exception, the rate should be assumed to be piecewise constant at the upper limit - i.e. the rectangle as you say.

Actually, the same argument can be made for the BHP curves (e.g., that they should be represented as  piecewise constant), they behave similar to the rate curves with discontinuities, so I am not sure I agree here. In any case, as already mentioned. The intent of the implementation under question is to have a measure of how different two simulators are. In this case, piecewise constant across the board is probably sufficient and makes for a very clean implementation. Hence I suggest no special treatment of BHP curves, unless it can be demonstrated to give significantly better measure of difference between BHP curves.

joakim-hove:
The point is that pressure is a _state_ variable - fundamentally different from a rate which is already a differential quantity. Spikes in the pressure is something different.

andlaus:
note that the piecewise constant values are a consequence of the need to be consistent with the time integration scheme: the implicit (and also explicit) Euler ODE discretization assumes that the function which it approximates is constant between the points in time which it considers (you could argue to use a better scheme like Runge-Kutta, but this would kill the match of the flow results with E100 which -- as far as I'm aware of -- is flow's main goal).

alfbr:
>  as far as I'm aware of -- is flow's main goal).

Different people probably have different goals. For me, E100 is simply a well established and well debugged simulator that is very valuable in providing reference solutions. As demonstrated by the MINPV and PINCH implementation in flow, we choose our own ways. If Runge-Kutta methods proved better, then at least I would be very interested in having it available in flow.

andlaus:
> Different people probably have different goals. [...]  If Runge-Kutta methods proved better, then at least I would be very interested in having it available in flow.

great! may I quote you on that next time we deviate from E100 for more than a few ppm? (be aware that changing such things will almost certainly kill comparability with E100 permanently, even if they improve things from a physical POV.)

alfbr:
> may I quote you on that next time we deviate from E100 for more than a few ppm? 

By all means :-)

CMakeLists_files.cmake
```diff
@@ -34,28 +40,29 @@ list (APPEND PUBLIC_HEADER_FILES
...
+        test_util/compareSummary.cpp					
     )
 
 # programs listed here will not only be compiled, but also marked for
 # installation
 list (APPEND PROGRAM_SOURCE_FILES
         test_util/initRegressionTest.cpp
         test_util/restartRegressionTest.cpp
-        test_util/summaryRegressionTest.cpp
+				test_util/compareSummary.cpp
```
pgdr: Please go through all whitespace and indentation strangeness.

opm/test_util/summaryComparator.cpp
```diff
@@ -45,6 +45,8 @@ SummaryComparator::SummaryComparator(const char* basename1, const char* basename
         this->keysShort = this->keys2;
         this->keysLong = this->keys1;
     }
+    
+
```
pgdr: try to minimize the number of lines you touch.  Adding unnecessary whitespace changes confuses the reviewers.

opm/test_util/summaryComparator.cpp
```diff
@@ -136,7 +138,7 @@ void SummaryComparator::getDeviation(size_t refIndex, size_t &checkIndex, Deviat
         return;
     }
     else if((*referenceVec)[refIndex]<(*checkVec)[checkIndex]){
-        double value = SummaryComparator::unitStep((*checkDataVec)[checkIndex-1]);
+        double value = SummaryComparator::unitStep((*checkDataVec)[checkIndex]);//Must be a little careful here. Flow writes out old value first, than changes value. Say there should be a change in production rate from A to B at timestep 300. Then the data of time step 300 is A and the next timestep will have value B. Must use the upper limit.
```
pgdr: Try to limit the maximum length of the lines.  You don't have to press it into 80 characters, but 350 is a bit over the top.

opm/test_util/summaryComparator.cpp
```diff
@@ -204,7 +206,71 @@ const char* SummaryComparator::getUnit(const char* keyword){
...
+		std::cout << keyword << std::endl; 
+		ivar++;
+	    }
+	    else{
+		noMatchString.push_back(keyword);
+		ivar++;
+	    }
+	}
+    if(noMatchString.size() == 0){
+	std::cout << "No keywords were different" << std::endl;
```
pgdr, line 197: be consistent with the space/tab indentation.  I believe OPM prefers spaces, but you should definitely not mix spaces and tabs in a file (let alone inside one function).

opm/test_util/summaryComparator.hpp
```diff
@@ -123,6 +138,16 @@ class SummaryComparator {
         static Deviation calculateDeviations( double val1, double val2);
         static double median(std::vector<double> vec);
 
+
+
+        //! \brief Sets the private member variable printKeywords
+        //! \param[in] boolian Boolian value
```
pgdr: minor nitpick: it's written "boolean" (with an "e").

opm/test_util/summaryIntegrationTest.cpp
```diff
@@ -0,0 +1,417 @@
...
+		ivar++;
+		continue;
+	    }
+	}  
+    
+	for (size_t jvar = 0; jvar < stringlist_get_size(keysLong); jvar++){
+
+	    if (strcmp(keyword, stringlist_iget(keysLong, jvar)) == 0){ //When the keywords are equal, proceed in comparing summary files. 
+	
+		//	if(!checkUnits(keyword)){OPM_THROW(std::runtime_error, "For keyword " << keyword << " the unit of the two files is not equal. Not possible to compare.");} //Comparing the unit of the two vectors.
```
pgdr: Please remove uncommented code.

By the way, if this PR is work in progress, please write that in the title and in the description of the PR.

opm/test_util/summaryIntegrationTest.cpp
```diff
@@ -0,0 +1,417 @@
...
+    }
+    return totalSum; 
+}
+
+double IntegrationTest::integrateError(std::vector<double> *timeVec1, std::vector<double> *dataVec1, std::vector<double> *timeVec2, std::vector<double> *dataVec2){
+
+//When the data corresponds to a rate the integration will become a Riemann sum. This function calculates the Riemann sum of the error. The reason why a Riemann sum is used is because of the way the data is written to file. When a change occur(e.g. change of a rate), the data (value and time) is written to file, THEN the change happenes in the simulator, i.e. we will notice the change at the next step. Keep in mind that the summary vector is NOT a continous curve, only points of data (time, value). We have to guess what happenes between the data points, we do this by saying: "There are no change, the only change happenes at the data points.". As stated above, the value of this constant "heigth" of the rectangle corresponds to the value of the last time step. Thus we have to use the "right hand side value of the rectangle as height (*someDataVector)[ivar] and NOT (*someDataVector)[ivar-1] (which intuition is saying is the correct value to use).
```
pgdr: This must be a new OPM record for long lines :D  Almost 1k characters.

Please break the lines, and remember, you are allowed to use paragraphs.  Example:

``` c++
// When the data corresponds to a rate the integration will become a Riemann
// sum.  This function calculates the Riemann sum of the error.  The reason why
// a Riemann sum is used is because of the way the data is written to file.
// When a change occur (e.g. change of a rate), the data (value and time) is
// written to file, THEN the change happens in the simulator, i.e., we will
// notice the change at the next step.
//
// Keep in mind that the summary vector is NOT a continuous curve, only points
// of data (time, value).  We have to guess what happens between the data
// points, we do this by saying: "There are no change, the only change happens
// at the data points."  As stated above, the value of this constant "height" of
// the rectangle corresponds to the value of the last time step.  Thus we have
// to use the "right hand side value of the rectangle as height
//
// (*someDataVector)[ivar] instead of (*someDataVector)[ivar-1]
//
// (which intuition is saying is the correct value to use).
```

opm/test_util/summaryRegressionTest.cpp
```diff
@@ -31,6 +31,8 @@ void RegressionTest::getRegressionTest(){
         OPM_THROW(std::invalid_argument, "Different ammont of keywords in the two summary files.");
     }
 
+    if(printKeyword){printKeywords();}
```
pgdr: If you insist on using one line, you may drop the braces (and add some spacing).

opm/test_util/summaryIntegrationTest.cpp
kristfho: Thanks, have updated the title of the PR.

kristfho: The reason why I have commented this code out is that flow does not write out which unit the keyword vector has. I was thinking this could be commented until flow writes out units, and than be uncommented. What do you think?

opm/test_util/summaryIntegrationTest.cpp
kristfho: Haha, yey, new record!

I used your suggestion in the code, thanks!

opm/test_util/summaryIntegrationTest.cpp
pgdr: As long as it's _work in progress_ you may keep it, but we prefer not having code that's out-commented.  It's better to just write in plain English what you mean.

kristfho:
Thanks for the comments, @pgdr. 
I have fixed most of them now.

joakim-hove:
Yes - but have you tested the totals from a restart ed Eclipse run? I know they will not be correct when using Flow - but thougth maybe eclipse for it right?

kristfho:
I will check first ting in the morning, @joakim-hove .

kristfho:
The restart file is set to start at timestep 120. 

This is what I get for an rate when doing an eclipse run 

```
Time          Ref data      Check data
0             0             0
4             1500          1500
10            1500          1500
...           ...           ...
110           100           100
111                         100
114                         100
120           100           100
130           100           100
```

And for the total: 

```
Time          Ref data      Check data
0             0             0
4             6000          6000
10            15000         15000
...           ...           ...
110           151000        151000
111                         151100
114                         151400
120           152000        152000
130           153000        153000
```

I'm not sure how to interpret this. I find it strange that both of the summary vectors have values from 0 - 120.

joakim-hove:
1. If a case has been restarted a pointer to the initial case is embedded in the summary results. The `ecl_sum` loader will look for and load this case - it is possible to tell it not to.
1. This shows that ECLIPSE gets the totals right even for a restarted case. My guess is that the previous totals are read from an undocumented keyword in the restart file.

kristfho:
This PR is now included in PR #70 . Closing now.

PR #77, opened 2016-08-18
WIP: Optionally write restart solution fields with double precision

joakim-hove:
This PR is semi-finished; but before merging is an option I would like to make a PR (and test it!) for opm-simulators as well.

### DIFF
--- a/opm/output/eclipse/EclipseReader.cpp
+++ b/opm/output/eclipse/EclipseReader.cpp
@@ -18,6 +18,7 @@
 */
 
 #include <ert/ecl/ecl_file.h>
+#include <ert/ecl/ecl_kw.h>
 #include <ert/ecl/EclKW.hpp>
 #include <ert/util/ert_unique_ptr.hpp>
 
@@ -36,6 +37,19 @@
 namespace Opm {
 namespace {
 
+    std::vector<double> double_vector( const ecl_kw_type * ecl_kw ) {
+        size_t size = ecl_kw_get_size( ecl_kw );
+
+        if (ecl_kw_get_type( ecl_kw ) == ECL_DOUBLE_TYPE ) {
+            const double * ecl_data = ecl_kw_get_double_ptr( ecl_kw );
+            return { ecl_data , ecl_data + size };
+        } else {
+            const float * ecl_data = ecl_kw_get_float_ptr( ecl_kw );
+            return { ecl_data , ecl_data + size };
+        }
+
+    }
+
     inline data::Solution restoreSOLUTION( ecl_file_type* file,
                                            int numcells,
                                            const UnitSystem& units ) {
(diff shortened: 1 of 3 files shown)


opm/output/eclipse/EclipseReader.cpp
```diff
@@ -36,6 +37,23 @@
 namespace Opm {
 namespace {
 
+    std::vector<double> double_vector( const ecl_kw_type * ecl_kw ) {
+        size_t size = ecl_kw_get_size( ecl_kw );
+        std::vector<double> opm_data( size );
```
jokva:
```
const size_t size = ecl_kw_get_size( ecl_kw );
if( ecl_kw_get_type( ecl_kw ) == ECL_DOUBLE_TYPE ) {
    const auto* ptr = ecl_kw_get_double_ptr( ecl_kw );
    return { ptr, ptr + size };
}

[...]
```

Would save you some manual work, and is slightly clearer (in my opinion).

joakim-hove: > Would save you some manual work, and is slightly clearer (in my opinion).

I agree - tried something similar with no success; and then resorted to deep manual mode. I will copy your suggestion verbatim.

joakim-hove:
OK - I have now used used this in opm-simulators (no PR yet - might never come?) and 
the last 25 elements of pressure field in the `SPE1CASE2` looks like this:

**kw1: No restart**:

``` txt
    3920.4449     3871.0008     3816.1738     3761.7514     3720.9508 
    4031.3357     4022.8485     4006.0708     3979.9608     3943.7766 
    3896.4146     3835.0654     3761.7514     3675.4381     3588.8810 
    4026.7181     4018.0229     4000.5362     3973.2218     3934.2327 
    3882.3820     3812.2036     3720.9508     3588.8810     3366.3248 
```

**kw2: With restart**

``` txt
    3920.4047     3870.9589     3816.1334     3761.7119     3720.9110 
    4031.2992     4022.8118     4006.0337     3979.9236     3943.7381 
    3896.3749     3835.0246     3761.7119     3675.3980     3588.8394 
    4026.6812     4017.9858     4000.4994     3973.1864     3934.1934 
    3882.3417     3812.1631     3720.9110     3588.8394     3366.2732
```

If we calculate `d = (kw1 - kw2) / kw1` we get avg(d) ~ 1e-5. For all that is worth the pressure is _consistently_ higher in the no restart case.

atgeirr:
While 1e-5 _may_ be not enough to worry, we did expect it to produce more or less identical results, right? What is the difference like just a single step after the restart?

Edit: also, what was this like with the float-based restart?

joakim-hove:
> While 1e-5 may be not enough to worry, we did expect it to produce more or less identical results, right? 

Yes!

> What is the difference like just a single step after the restart?

That was very interesting. The list below shows the `<(kw1 - kw2) / kw1>` for each step after the restart: 

``` txt:
 0.0002852
 0.0002814
 0.0002648
 0.0002401
 0.0002153
 0.0001965
 0.0001829
 0.0001736
 0.0001670
 0.0001623
 0.0001587
 0.0001559
 0.0001536
 0.0001520
 0.0001500
 0.0001479
 0.0001456
 0.0001448
 0.0001440
 0.0001421
 0.0001382
 0.0001342
 0.0001300
 0.0001258
 0.0001216
 0.0001177
 0.0001133
 0.0001088
 0.0001040
 0.0000994
 0.0000945
 0.0000893
 0.0000839
 0.0000711
 0.0000624
 0.0000516
 0.0000397
 0.0000316
 0.0000270
 0.0000270
 0.0000309
 0.0000365
 0.0000419
 0.0000448
 0.0000449
 0.0000433
 0.0000409
 0.0000380
 0.0000349
 0.0000320
 0.0000289
 0.0000259
 0.0000230
 0.0000205
 0.0000181
 0.0000161
 0.0000144
 0.0000128
 0.0000114
 0.0000102
```

I'll double check tomorrow that I am not messing up the comparison; but the difference actually decays. If this is indeed correct my hunch would be that the first step is wrong in some way, and then the state asymptotically approaches a boundary conditions controlled state.

joakim-hove:
OK - the restart problems are certainly _not_ only related to the `double -> float -> double` conversions. In that sense this PR has served it's purpose; I would kind of like to merge it anyway - but I will leave that to @jokva to decide.

PR #79, opened 2016-08-20
Restart: write properties

joakim-hove:
Added the ability for the restart writer to write arbitrary fields in the restart section. Example of such a field would be the relperms KRO, KRG and KRW and the fluxes. The fields must be calculated and provided by calling scope - i.e. flow in this case.

If not exactly _work in progress_ this is on a strictly 'good enough' basis; tricky balance to find. There is a minor argument reordering which must be fixed in opm-simulators: https://github.com/OPM/opm-simulators/pull/795


jokva:
This is fine I suppose, at least until we get generalised configurable restart files. Whenever you feel downstream is ready, press the button.

PR #82, opened 2016-08-25
Summary: write region pressure keyword RPR

joakim-hove:
With this PR we get support for the RPR:xxx keywords in the summary file. The patch works, but it is certainly the first of the region keywords, and some new patterns for supporting the region keywords should probably emerge - this is very early days. In particualar some argument lists are embarassingly long.

The EclipseWriter object will cache a map of { region_id : cells } for faster evaluation of the region sums. The underlying methods in opm-parser: https://github.com/OPM/opm-parser/pull/902 can probably also be used here: https://github.com/OPM/opm-simulators/pull/766?


jokva:
Planning on anything more here? It's merge-at-will at this point, unless you have more in the pipeline.

opm/output/eclipse/Summary.cpp
```diff
@@ -118,8 +118,11 @@ struct fn_args {
     const std::vector< const Well* >& schedule_wells;
     double duration;
     size_t timestep;
-    int active_index;
+    int  num;
     const data::Wells& wells;
+    const data::Solution& state;
+    const EclipseState* es;
```
jokva: `3DProperties`.

Then merge at will.

joakim-hove:
> Planning on anything more here?

Yes - close to pushing a shiny improved patch :grinning:

jokva:
Alright, then I'll have a second look once it's up and obviously merging is now disallowed.

joakim-hove:
OK - update pushed; now this code does at least not make me blush any longer.

jokva:
I'll have a look.

opm/output/eclipse/Summary.cpp
```diff
@@ -265,6 +269,23 @@ inline quantity duration( const fn_args& args ) {
     return { args.duration, measure::time };
 }
 
+quantity rpr(const fn_args& args) {
+    const auto& pair = args.regionCells.find( args.num );
+    if (pair != args.regionCells.end()) {
```
jokva: General tip/preference: negate the check and exit early to save indentation and to avoid having to maintain context and state while reading the function.

``` C++
if( exit_condition ) return;

/*
    don't have to wonder if state update here is needed after the block.
*/
```

Instead, now I have to read as this.

``` C++
if( condition ) {
   /* some state is modified */
} else {
   /* some other state modification */
}

/* is code here dependent on which code path was chosen earlier? */
```

By using the former, the latter question just doesn't surface. Also, it's in essence a short circuit and I strongly prefer when totally disjoint code path nodes don't share a child node, simply because I need to trace both paths to figure out what may or may not have happened.

Two cents etc. and not that big of a deal for such a small function.

YMMV.

opm/output/eclipse/Summary.cpp
```diff
@@ -443,6 +464,9 @@ static const std::unordered_map< std::string, ofun > funs = {
     { "FGLRH", div( production_history< Phase::GAS >,
                     sum( production_history< Phase::WATER >,
                          production_history< Phase::OIL > ) ) },
+
+    /* Region properties */
+    { "RPR" , rpr}
```
jokva: **If** you chose to amend to this patch, a handy tip:

A trailing comma in an init list like this is legal. If you add it here (as in it was in the case of FGLRH) then it won't show up in the diff next time we add a keyword because then we won't have to add a comma.

opm/output/eclipse/Summary.cpp
```diff
@@ -522,8 +546,9 @@ Summary::Summary( const EclipseState& st,
 
         /* get unit strings by calling each function with dummy input */
         const auto handle = funs.find( keyword )->second;
-        const std::vector< const Well* > dummy;
-        const fn_args no_args{ dummy, 0, 0, 0, {} };
+        const std::vector< const Well* > dummy_wells;
+        const std::unordered_map<int,std::vector<size_t>> dummy_cells;
+        const fn_args no_args{ dummy_wells, 0, 0, 0, {} , {}, dummy_cells , {} };
```
jokva, line 551: thumbs up.

opm/output/eclipse/Summary.cpp
joakim-hove: OK - that was certainly news to me.

joakim-hove:
OK - minor review comments addressed and force pushed.

jokva:
Splendid. Go when green.

PR #83, opened 2016-08-29
Constant rename

joakim-hove:
Premature mismerge here: https://github.com/Ensembles/ert/pull/1214


jokva:
Trivial change, go whenever.

PR #84, opened 2016-09-01
Output writer cell data

babrodtk:
Enables outputting cell data to UNRST files.

See also sibling pull requests in OPM/opm-simulators#805 and OPM/opm-core#1060


opm/output/Cells.hpp
```diff
@@ -32,6 +32,10 @@ class SimulationDataContainer;
 namespace data {
 
     struct CellData {
+        CellData(std::string name, UnitSystem::measure dim, std::vector<double> data)
```
joakim-hove: I guess the data could be `std::vector<double>&` - and the name as well?

bska: Or, the members should be initialised using `std::move()`.

jokva: Since it's being copied anyway passing the arugments by value is fine (the copy might be elided this way). Clearly communicates what we're doing too, which is a bonus in my opinion.

atgeirr: Since this is a simple struct you can initialize it using `CellData{ name, dim, data }` even without this constructor, isn't that right?

babrodtk: In theory, I believe you are correct, but I got a compilation error using my gcc-5.4.0 without this constructor.

jokva: > Since this is a simple struct you can initialize it using CellData{ name, dim, data } even without this constructor, isn't that right?

Yes.

atgeirr: > I got a compilation error

Interesting, I wonder why?

In any case adding the constructor is a good enough workaround. I agree we might want to use an rvalue-reference for `data` though.

jokva: > Interesting, I wonder why?

Did it wrong, probably. It shouldn't be needed.

```
$ cat struct.cpp; ./a.out 
#include <vector>
#include <string>
#include <iostream>

struct foo {
    std::string x;
    std::vector< int > y;
};

int main() {
    std::string x = "foo";
    std::vector< int > y( 10, 4 );

    foo f { x, y };

    std::cout << "x: " << f.x << std::endl;
    std::cout << "y: ";
    for( auto e : f.y )
        std::cout << e << " ";
    std::cout << std::endl;
}

x: foo
y: 4 4 4 4 4 4 4 4 4 4 

```

> In any case adding the constructor is a good enough workaround. I agree we might want to use an rvalue-reference for data though.

I presume you mean `std::move` in the init list? Then yes.

Regardless, if you remove the constructor and brace-initialise this problem goes away.

jokva: Nevermind, wrong repo.

jokva: > Interesting, I wonder why?

My guess would be he used a vector or some other container of these and tried to use `emplace_back`. Using `struct foo` from my example above:

```
std::string x; std::vector< int > y;
vec foo;
vec.emplace_back( x, y ); // won't compile
vec.emplace_back( { x, y } ); // won't compile
vec.emplace_back( foo{ x, y } ); // compiles
```

edit: extended the example for completeness.

opm/output/OutputWriter.hpp
```diff
@@ -94,7 +99,8 @@ class OutputWriter {
                                 time_t current_posix_time,
                                 double seconds_elapsed,
                                 data::Solution reservoirState,
-                                data::Wells,
+                                data::Wells wellState,
+                                const std::vector<data::CellData> cellData,
```
atgeirr, line 103: It turns out that this change is not needed, as this abstract class is not used anymore at all! (There is an unneeded include in opm-simulators, that is all).

opm/output/Cells.hpp
babrodtk: the emplace_back example you show is just what caused the problem, yes.

atgeirr:
Can this be merged? It is required for downstream PRs (forgot about it when making the similar request in opm-parser, sorry for being slow...).

Removing the unused header is best done separately.

jokva:
No change needed if downstream uses brace initialisation.

atgeirr:
It does not, uses traditional emplace_back instead.

joakim-hove:
I must admit I have a slightly iffy feeling for adding a constructor to the `struct` - I prefer one of:
1. Full blown C++ class with data encapsulation and methods and so on.
2. Fully public C style struct which  only aggregates some related data.

So if you could just do `simProps.emplace_back(CellData{ "Key", unit , data });` in opm-simulators I would prefer that. If you consider that is a bad solution I will certainly merge the constructor, but then please also add a default constructor so that 

``` C++
CellData cd;
```

still compiles.

babrodtk:
Changed downstream code so that it does not require these changes.

PR #87, opened 2016-09-07
Outputwriter: use EclipseGrid

joakim-hove:
- The outputwriter will now take a a separate EclipseGrid instance as
  input argument, it is assumed that calling scope has already made sure
  ACTNUM and ZCORN are correct.
- All active/inactive cell mappings are based on the grid argument, the
  naked int\* with global / active cell mappings has been completely
  removed.

Upstream: https://github.com/OPM/opm-parser/pull/915
Downstream: https://github.com/OPM/opm-simulators/pull/814

### DIFF
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -60,26 +60,6 @@
 namespace Opm {
 namespace {
 
-// throw away the data for all non-active cells and reorder to the Cartesian logic of
-// eclipse
-template <typename T>
-std::vector<T> restrictAndReorderToActiveCells(const std::vector<T> &flow_data,
-                                               int numCells,
-                                               const int* compressedToCartesianCellIdx)
-{
-    std::vector<T> eclData;
-    eclData.resize( numCells );
-
-    if (!compressedToCartesianCellIdx || (flow_data.size() == static_cast<size_t>(numCells)))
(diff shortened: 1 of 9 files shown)


opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -853,15 +793,29 @@ EclipseWriter::EclipseWriter( std::shared_ptr< const EclipseState > es,
...
-            this->impl->regionCells.emplace( region_id , fipnum.cellsEqual( region_id , this->impl->compressedToCartesian ));
+            this->impl->regionCells.emplace( region_id , fipnum.cellsEqual( region_id , this->impl->grid ));
     }
 }
 
+
+EclipseWriter::EclipseWriter(std::shared_ptr< const EclipseState > es, const EclipseGrid& grid)
+    : impl( new Impl( es , grid) )
+{
+    init();
```
jokva: This needs to go away.

A better idea is actually to only pass the `EclipseGrid` argument by value (since it's being copied anyway) and not overload the constructor. It removes the needs for a common `init`, makes it obvious what's going on, and caller can choose to move if he wants to.

jokva:
_Great_ work and patch! Glad to see the inactive/active mappings go away. Looking forward to seeing this merged.

joakim-hove:
OK - merging this now; the combined Jenkins build at opm-simulators is green.

PR #90, opened 2016-09-14
Restructure test utilities and add integration test

akva2:
- reduce number of test applications and rename
- add integration test utilities

Replaces #70


opm/test_util/EclFilesComparator.cpp
```diff
@@ -20,173 +20,105 @@
...
+    return numCells1;
 }
 
 
 
-void ECLFilesComparator::deviationsForCell(double val1, double val2, const std::string& keyword, int occurence, int cell, bool allowNegativeValues) {
+template <typename T>
+void ECLFilesComparator::printValuesForCell(const std::string& keyword, int occurrence1, int occurrence2, size_t cell, const T& value1, const T& value2) const {
     int i, j, k;
     ecl_grid_get_ijk1(ecl_grid1, cell, &i, &j, &k);
```
joakim-hove, line 64: The call to `ecl_grid_get_ijk1()` is probably wrong - the `ecl_grid_get_ijk1()` expects the cell index to be in the range `[0,nx*ny*nz)` - whereas it looks to me like `cell`will be in the range `[0,nactive)`. The function: `ecl_grid_get_ijk1A()` will take an _active_ index instead.

bska: > The call to `ecl_grid_get_ijk1()` is probably wrong

We actually have no way of knowing that.  The `cell` is in the range `[0, getEclKeywordData(...))` and `getEclKeywordData()` returns `ecl_kw_get_size()` of a named ECL keyword.

joakim-hove: > We actually have no way of knowing that.

As I said - _probably_.

bska: > > We actually have no way of knowing that.
> 
> As I said - probably.

True.  On the other hand, since we do not know what function to call, that is a sign that there is a possibly a piece of information missing at the call site.

akva2:
i am experiencing machine-dependent output in the INTEHEAD field in unrst files.

ref http://paste.ubuntu.com/23181477/

the 'ref' data is what is stored in opm-data, and what i get on my machine.
the 'jnk' data is what i get on the jenkins build box.

it seems to be a rounding-issue for a date, at least that's what my uninformed guess would point to. any cluesticks  you can offer..

joakim-hove:
~~Ok - two bottles of bear that this is timezone/locale issue, will give you more.~~

Some of the elements in the `INTEHEAD` keyword are documented [here](https://github.com/Ensembles/ert/blob/master/libecl/include/ert/ecl/ecl_kw_magic.h#L43). Looks like the reference solution has embedded the date 31.th of January 2016, whereas the Jnk solution has 1.st of February 2016.

My initial hunch was that this was due to computers in different timezones, but `localtime()` should not really be involved at any stage in this, so that is probably not the case. My next hunch would be that you are comparing solution blocks from two different time steps. To test I would:
1. Make sure no stale files are lying around.
2. Remove the `UNIFIN` and `UNIFOUT` keywords from the deck; or alternatively use the ert utility `ecl_unpack.x` to split the UNRST file. Then you will have a set of `.Xnnnn` files and you are guaranteed to not mess up the report steps when comparing.

akva2:
it passes fine on my machine, same code, same files. there is no stale files, output is to a fresh directory on every run. exactly the same happens with split output. i traced it to 

```
#0  __util_set_timevalues (t=1454284799, sec=0x0, min=0x0, hour=0x0, mday=0x7fffffffc9e4, month=0x7fffffffc9ec, year=0x7fffffffc9e8) at /home/akva/kode/opm/sibling/ert/libert_util/src/util.c:3086
#1  0x00007ffff61de347 in util_set_date_values (t=1454284799, mday=0x7fffffffc9e4, month=0x7fffffffc9ec, year=0x7fffffffc9e8) at /home/akva/kode/opm/sibling/ert/libert_util/src/util.c:3109
#2  0x00007ffff6669ab3 in ecl_util_set_date_values (t=1454284799, mday=0x7fffffffc9e4, month=0x7fffffffc9ec, year=0x7fffffffc9e8) at /home/akva/kode/opm/sibling/ert/libecl/src/ecl_util.c:1538
#3  0x00000000012edad4 in Opm::(anonymous namespace)::Restart::writeHeader (this=0x7fffffffc9a0, stepIdx=8, rsthead_data=0x7fffffffc9e0) at /home/akva/kode/opm/sibling/opm-output/opm/output/eclipse/EclipseWriter.cpp:238
```

where an util_localtime derived time is used.

joakim-hove:
Ok - so it was actually a localtime issue.

akva2:
yes. not sure what to do now.

joakim-hove:
I will have a look later, what is the operating system details (including LOCALE) for the servers involved.

akva2:
my box;

```
date -> to. 15. sep. 13:20:29 +0200 2016 
cat /etc/timezone -> Europe/Oslo
locale ->
LANG=nb_NO.UTF-8
LANGUAGE=en_US
LC_CTYPE="nb_NO.UTF-8"
LC_NUMERIC="nb_NO.UTF-8"
LC_TIME="nb_NO.UTF-8"
LC_COLLATE="nb_NO.UTF-8"
LC_MONETARY="nb_NO.UTF-8"
LC_MESSAGES="nb_NO.UTF-8"
LC_PAPER="nb_NO.UTF-8"
LC_NAME="nb_NO.UTF-8"
LC_ADDRESS="nb_NO.UTF-8"
LC_TELEPHONE="nb_NO.UTF-8"
LC_MEASUREMENT="nb_NO.UTF-8"
LC_IDENTIFICATION="nb_NO.UTF-8"
LC_ALL=
```

jenkins box:

```
date -> Thu Sep 15 11:20:55 UTC 2016 
cat /etc/timezone -> Etc/UTC
locale ->
LANG=nb_NO.UTF-8
LANGUAGE=
LC_CTYPE="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_COLLATE="C"
LC_MONETARY="C"
LC_MESSAGES="C"
LC_PAPER="C"
LC_NAME="C"
LC_ADDRESS="C"
LC_TELEPHONE="C"
LC_MEASUREMENT="C"
LC_IDENTIFICATION="C"
LC_ALL=C
```

joakim-hove:
Env variables `TZ` and `LOCALE`?

akva2:
empty on both boxes.

akva2:
added contents of /etc/timezone.

joakim-hove:
OK; just to be absolutely certain I am not barking up the wrong tree: Can you confirm that the `rsthead->sim_time` argument has the same value [here](https://github.com/OPM/opm-output/blob/master/opm/output/eclipse/EclipseWriter.cpp#L235)?

akva2:
yes. with

``` diff
diff --git a/opm/output/eclipse/EclipseWriter.cpp b/opm/output/eclipse/EclipseWriter.cpp
index 0628013..7c3d09e 100644
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -236,6 +236,7 @@ public:
                                 &rsthead_data->day,
                                 &rsthead_data->month,
                                 &rsthead_data->year );
+      std::cout << "day " << rsthead_data->day << " " << rsthead_data->month << " " << rsthead_data->year << " " << rsthead_data->sim_time << std::endl;
       ecl_rst_file_fwrite_header( this->rst_file.get(), stepIdx, rsthead_data );

     }
```

on my box;

```
day 1 2 2016 1454284799
```

on the jenkins box;

```
day 31 1 2016 1454284799
```

the latter corresponds to the date printed for the _next_ time step in the simulation loop, i.e. the 'day' output comes before on the terminal. it prints the same on both machines.

bska:
``` sh
$ TZ=Europe/Oslo date --date='@1454284799'
Mon Feb  1 00:59:59 CET 2016
$ TZ=Etc/UTC date --date='@1454284799'
Sun Jan 31 23:59:59 UTC 2016
```

akva2:
yup.

joakim-hove:
OK - I guess anyone who has followed this has a quite clear idea of what has happened. But anyway my take on the situation:
1. When writing a restart file the true time as posix `time_t` variables is passed to the output routine. The `time_t` variable is converted to the triplet (Year, Month, Day) - and that is saved in the `INTEHEAD` keyword in the restart files.
2. The conversion has been performed with the function `localtime()` which is timezone aware.
3. The Jenkins build server and the typical developer machines have been in different timezones, and the result is that the files are different.

To fix this:
1. The reference case must be updated - since the temporal resolution in `INTEHEAD` is days it requires some "bad luck" to hit points of conflict, so hopefully not all cases must be updated.
2. Convert all the time handling in ert to use UTC. The process has started: https://github.com/Ensembles/ert/pull/1229 - but this will be a flag day, and take some time.

joakim-hove:
OK - this should fix it:  Ensembles/ert#1229. Will come with an updated case later today.

akva2:
great!

akva2:
i can confim the issue is fixed. i will have to wait for the opm-output test update PR you mention before i proceed to have valid hashes for the opm-data update.

joakim-hove:
A bit uncertain what you mean here. The only thing remaining - as I see it - is to create updated test data. I am happy to make the required PR to opm-data, but will not manage that before later today/tomorrow.

akva2:
yeah, that part was more a FYI for @alfbr than a push in your direction. point is I list hashes in the opm-data commit messages, and i can't give a valid hash for a working, tested opm-output (i have local hacks to make it build..) until the tests are in place. there's slightly more than updating test data involved, there are some calls to the ert time function that disappeared with https://github.com/Ensembles/ert/pull/1229 see travis build log for instance

akva2:
tested through http://80.240.129.131:8080/job/opm-simulators-PR-builder/112/ ready for merge from my side.

akva2:
since this only touches code i am supposed maintain, i'm merging.

PR #91, opened 2016-09-14
Summary output of RWIP

joakim-hove:


joakim-hove:
Unless .... I'll merge this tomorrow.

jokva:
No objections.

opm/output/eclipse/Summary.cpp
```diff
@@ -286,6 +286,25 @@ quantity rpr(const fn_args& args) {
...
+    if (pair == args.regionCells.end())
+        return { 0.0 , measure::volume };
+
+    double RWIP = 0;
+    const std::vector<double>& swat = args.state[data::Solution::key::SWAT];
+    const auto& cells = pair->second;
+
+    for (auto cell_index : cells) {
+	size_t global_index = args.grid.getGlobalIndex( cell_index );
+        RWIP += swat[cell_index] * args.grid.getCellVolume( global_index );
```
pgdr: Should also multiply by `porv[cell_index]`

joakim-hove: > Should also multiply by porv[cell_index]

Well spotted  - thank you; amend pushed.

opm/output/eclipse/Summary.cpp
```diff
@@ -286,6 +287,32 @@ quantity rpr(const fn_args& args) {
...
+
+
+    double RWIP = 0;
+    const std::vector<double>& swat = args.state[data::Solution::key::SWAT];
+    const auto& cells = pair->second;
+    const auto& properties = args.ecl_state->get3DProperties();
+    const auto& porv_data = properties.getDoubleGridProperty("PORV").getData();
+
+    for (auto cell_index : cells) {
+	size_t global_index = args.grid.getGlobalIndex( cell_index );
```
pgdr, line 307: somehow a tab has snuck in

pgdr:
Ready to merge after the tab has been replaced.

alfbr:
I am afraid you need to include the water formation volume factor, assuming that in place numbers are at surface conditions. For the Norne case, you need to calculate it from the water compressibility in the PVTW table. Since these numbers already are calculated in the simulator, and even done in parallel there, consider doing this on the simulator side.

jokva:
We already compute original and current water/oil/gas in place in the simulator. Can we just pass this result as-is to output and just read that, or do I miss something? 

Line ~274 in SimulatorBase_Impl.hpp. If these numbers are the same we want to print it looks like we'll get ROIP, RGIP, RWIP, FWIP, FGIP and FOIP for free, which is nice.

alfbr:
> We already compute original and current water/oil/gas in place in the simulator. Can we just pass this result as-is to output and just read that, or do I miss something?

I believe so, and that is what I am asking to consider.

opm/output/eclipse/Summary.cpp
```diff
@@ -286,6 +287,32 @@ quantity rpr(const fn_args& args) {
...
+        RWIP += swat[cell_index] * porv_data[global_index];
```
andlaus, line 308: you probably also have to multiply this with B_W. (at least if RWIP is given in terms of surface volume, which IMO is the only sensible option.)

Also, I assume that "swat" corresponds to the water saturation given by the solution, not to the values given by the SWAT ECL property. (there is potential for confusion.)

andlaus (commented):
IMO, this looks good once it has been clarified if the multiplication with B_W must be done or not.

Question: `R[OG]IP` are not yet supported? (I haven't folllowed this very closely.)

andlaus:
> We already compute original and current water/oil/gas in place in the simulator. Can we just pass this result as-is to output and just read that, or do I miss something?

for the current `flow` in principle yes, in practice this probably implies some annoying opm-output API changes? maybe it is better to move the actual computation to the simulator and limit opm-output to a querying API for the requested quantities and to writing the results to disk.

jokva:
> for the current flow in principle yes, in practice this probably implies some annoying opm-output API changes?

Possibly a change to the input (my guess would be `data::Solution`) structures, yes. No change unless you want to output fluid-in-place though.

> maybe it is better to move the actual computation to the simulator and limit opm-output to a querying API for the requested quantities and to writing the results to disk.

Isn't this what I suggested?

andlaus:
> Isn't this what I suggested?

I suppose you did. it seems that I was just too stupid to notice, sorry...

jokva:
No problems mate, just wanted to make sure I didn't miss a detail.

joakim-hove:
> Since these numbers already are calculated in the simulator, and even done in parallel there, consider doing this on the simulator side.

We will do that; there are some unrelated refactorings in this patch I would like to salvage, but apart from that we will take this from the simulator.

jokva:
> We will do that; there are some unrelated refactorings in this patch I would like to salvage, but apart from that we will take this from the simulator.

Alright, we'll take a look tomorrow then, yeah?

joakim-hove:
OK - nothing worth keeping here. Closing.

PR #92, opened 2016-09-19
Use ecl_util_make_date() to create time_t value.

joakim-hove:
This fixes a fallout after: https://github.com/Ensembles/ert/pull/1229

Requires: https://github.com/Ensembles/ert/pull/1229


joakim-hove:
This fixes the current build failure - will self merge.

PR #93, opened 2016-09-21
Fix restart for shut wells and no wells

totto82:
Needs OPM/opm-parser#925


joakim-hove:
Given the format of the ECLIPSE file we need to write restart information for all wells for every timestep, possibly just outputting dummy values for not-yet-defined wells. This is not correctly handled in opm-output - and we must fix it. But the approach taken in this patch is unfortunately not viable - sorry!

atgeirr:
> we need to write restart information for all wells for every timestep, possibly just outputting dummy values for not-yet-defined wells

That is interesting, I guess the requirement for doing it even for not-yet-defined wells means that any proper solution will not require the simulator side to provide dummy output for currently-shut wells?

joakim-hove:
> I guess the requirement for doing it even for not-yet-defined wells means that any proper solution will not require the simulator side to provide dummy output for currently-shut wells?

No - the output layer should be able to handle that.

blattms:
Sorry for highjacking this discussion.
What happens/should happen if for eclipse a well is open and has open perforations but the simulators decides that these perforations should be shut (e.g. because the cell is inactive or some limiter `DynamicListEconLimited`  tells it to shut it.)?

joakim-hove:
> What happens/should happen if for eclipse a well is open and has open perforations but the simulators decides that these perforations should be shut (e.g. because the cell is inactive or some limiter DynamicListEconLimited tells it to shut it.)?

I don't the full details; but for the context of this PR the answer is: "The well - with the appropriate settings , should be written to the restart file".

jokva:
> That is interesting, I guess the requirement for doing it even for not-yet-defined wells means that any proper solution will not require the simulator side to provide dummy output for currently-shut wells?

Correct, but I don't think we'd mind if the simulator filled some closed wells with garbage and submitted it for output (if that makes it easier to generalise for you). I doubt that's the case though, and it should be pretty easy for output to do that.

PR #94, opened 2016-09-22
Output SSOL to restart file if present

totto82:
Needs OPM/opm-core#1069

### DIFF
--- a/opm/output/Cells.hpp
+++ b/opm/output/Cells.hpp
@@ -50,6 +50,7 @@ namespace data {
             SGAS,
             RS,
             RV,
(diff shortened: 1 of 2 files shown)


joakim-hove:
OK - this is fine and can be merged. The dependency between this and OPM/opm-core#1069 is not entirely clear to me, but I will leave it to e.g. @atgeirr to merge this when OPM/opm-core#1069 is going in.

totto82:
This could safely be merged. Its only depend on OPM/opm-core#1069 in the sense that it would not be used if the PR in opm-core is merged.

PR #95, opened 2016-09-22
[WIP] fixed: store well control states and perforation rates for restart

akva2:
these are needed for a perfect restart. well controls is necessary because state transitions affects the model. the perforation rates might be recomputable but due to order in the current simulators, they are not available for first assembly.

this is WIP and opened early because i need input from those with the knowledge. i have only compile-fixed the tests, some of them fail.


joakim-hove:
This code is quite simple - and mostly a slight alteration of the datastructures in `WellState` - so I guess the current PR is fine. My only question: The current PR adds the control values to the `OPM-XWEL` keyword, but are we guaranteed that the current control _mode_ is also serialized? I fear we will have to introduce a `OPM-IWEL` to handle that.

akva2:
i don't understand what you mean, which is likely just my lacking understanding. what is stored is the mode of the well, which is encoded as an integer for each well, cast to/from double on write/read to not introduce another field in the file. you believe there are data missing? can you point to the field you think is missing in the WellStateFullyImplicitBlackoil.hpp?

atgeirr:
> integer for each well, cast to/from double

The output format supports integer fields natively, so we could avoid this, as @joakim-hove alludes to the XWEL stores floating-point data for the wells while IWEL stores integer data.

Actually, the most robust thing to store might be the string name of the control type, i.e. LRAT, ORAT, BHP etc., that may be too much hassle though.

akva2:
sure, if that's all implied i follow, i just did the quickest approach. unless you mean to change your enum imho string serialization won't add much robustness as this is internal simulator state you fully control, and nothing but flow_xxx will ever write the field.

joakim-hove:
> [....], which is likely just my lacking understanding.

Or mine .... 

> what is stored is the mode of the well, which is encoded as an integer for each well, cast to/from double on write/read to not introduce another field in the file. you believe there are data missing? 

OK; I really think integer values should be stored as integers, i.e. in the `IWEL` vector. Since we do not know where in the `IWEL`vector this should be stored, I suggest we add a new `OPM-IWEL` keyword.

What I _thought_ was stored in the `cnt` vector was something like the current solution of the various control constraints.

akva2:
right, it has a slightly more verbose name in the prototype. i tried to stick to the code style used.

if you want an OPM_IWEL you shall have an OPM_IWEL.

akva2:
there you go. also fixed tests.

jokva:
I've had a chat with Joakim and we've concluded that we need to fix the underlying architectural issues before trying to get a perfect restart in place. We'll use this work as a guideline once the proper structures are in place (with due credit), but I don't think we'll even look to merge this as-is.

The plan is to remove the free bhp/temperature/pressure vectors in Wells and incorporate them into the `Well` struct, giving the properties names and labels.

Do you see any problems with this?

akva2:
not at all. i have put this on top of the solvent stuff for now.

jokva:
Alright, I'll let you all know (with a PR probably) when things are shaping up.

joakim-hove:
> I've had a chat with Joakim and we've concluded that we need to fix the underlying architectural issues before trying to get a perfect restart in place.

OK - I have had a subsequent chat with @alfbr - and this might have come out a bit harsh. The point as we see it is the following:

1, The current patch identifies and fixes problems related to performing perfect restart :white_check_mark: 
2. There have been issues identified with restarting cases where not all wells are present/open. Fixing this properly will require some refactoring of the opm-output code, which will guaranteed be in conflict with the current PR.
3. The current PR must be merged together with https://github.com/OPM/opm-core/pull/1070[1]

Jørgen and myself exepcted that within the time https://github.com/OPM/opm-core/pull/1070 was ready to merge the current PR would already be stale, due to the refactorings required to handle closed wells in restart. However; if/when https://github.com/OPM/opm-core/pull/1070 is ready to be merged quite soon, we can merge this also. Opm-core maintainers should indicate their preference, and we will act accordingly.

[1]: Regarding https://github.com/OPM/opm-core/pull/1070 there seems to be some ambiguity - I assume https://github.com/OPM/opm-core/pull/1071 is an alternative, or in conflict with https://github.com/OPM/opm-core/pull/1070?

atgeirr:
I think OPM/opm-core#1070 shows what is wrong, but the fix is not general enough (I intend to provide a more general fix Monday probably). Also, we are moving the file to opm-simulators, so there is a small chain of events that need to happen before what 1070 eventually turns into is merged.

jokva:
I've implemented OPM_IWEL support for registering well controls in the restart file in https://github.com/OPM/opm-output/pull/98

opm-simulators isn't quite ready to actually read this data yet (it's provided for them, but there's no output-to-WellStateFullyImplicitBlackoil (or whatever it's called)), but the state->output conversion seems to work.

jokva:
Closing as this feature has been implemented in master.

PR #97, opened 2016-09-26
Added boolean flag for writing cell data to restart files

babrodtk:


joakim-hove:
OK - this looks fine; but please also _use_ the new bool [here](https://github.com/OPM/opm-output/blob/master/opm/output/eclipse/EclipseWriter.cpp#L744).

babrodtk:
@joakim-hove Done (I hope :) )

PR #98, opened 2016-09-26
WIP: Redesign xwel writing

jokva:
Redesign the writing of OPM_XWEL with a corresponding change in `data::Wells`.

@totto82 Please help me test this.

### DIFF
--- a/opm/output/data/Wells.hpp
+++ b/opm/output/data/Wells.hpp
@@ -83,41 +83,19 @@ namespace Opm {
         using active_index = size_t;
         active_index index;
         Rates rates;
+        double pressure;
+        double reservoir_rate;
     };
 
     struct Well {
         Rates rates;
         double bhp;
         double thp;
+        double temperature;
         std::map< Completion::active_index, Completion > completions;
     };
 
-    struct Wells {
-        using value_type = std::map< std::string, Well >::value_type;
-        using iterator = std::map< std::string, Well >::iterator;
-
-        inline Well& operator[]( const std::string& );
-        inline Well& at( const std::string& );
-        inline const Well& at( const std::string& ) const;
-        template< typename... Args >
-        inline std::pair< iterator, bool > emplace( Args&&... );
-
-        inline Wells() = default;
-        inline Wells( std::initializer_list< value_type > );
-        inline Wells( std::initializer_list< value_type >,
-                      std::vector< double > bhp,
-                      std::vector< double > temperature,
-                      std::vector< double > wellrates,
-                      std::vector< double > perf_pressure,
-                      std::vector< double > perf_rates );
-
-        std::map< std::string, Well > wells;
-        std::vector< double > bhp;
-        std::vector< double > temperature;
-        std::vector< double > well_rate;
-        std::vector< double > perf_pressure;
-        std::vector< double > perf_rate;
-    };
+    using Wells = std::map< std::string, Well >;
 
     /* IMPLEMENTATIONS */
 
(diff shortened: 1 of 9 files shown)


jokva:
To clarify, this should not break current restart support, but you won't be able to restart from files not generated with this patch applied.

@totto82 This should also fix your missing well issue.

totto82:
I have tested this PR together with its friends. And I am now not even able to run Model 2. I simply get a std::bad_alloc. My guess without looking at the code at all is that this is triggered by the fact that there is no wells in the model. Maybe a simple test case without wells in the first steps would be useful.

jokva:
Curious - when there are no wells, writing should default to a no-op. Do you have a trace for me?

blattms:
On Wed, Sep 28, 2016 at 12:14:11AM -0700, jokva wrote:

> Curious - when there are no wells, writing should default to a no-op.

Well, there certainly are wells. But at the beginning of the
simulation they are all shut.

There are even more pecularities that may arise and cause problems,
like open wells with allcompletions in inactive cells. See OPM/opm-core#1074

jokva:
Ok, I'm on it. Found a `bad_alloc` bug in the opm-simulators PR.

jokva:
@totto82 Mind updating simulators and giving it a spin?

totto82:
On it.

totto82:
The `bad_alloc` issue seems to be fixed. No I get 
`Program threw an exception: Input argument does not correspond to an active cell`
after some time when I run the model without restart and immediately when I try to restart.

jokva:
Ok, good. I have an idea on what's causing this: the mapping cell index -> active index is wrong somewhere (likely, as I don't have a proper overview on what's what in the simulator).

How do you perform the restart? Args to flow etc?

jokva:
Is the well_cells array in the wells struct not the active index of the completions of the respective wells?

jokva:
ping @atgeirr

jokva:
Ok, so I have a break point at where the exception is thrown. It's the cell (71, 145, 21) that's inactive as we attempt to serialize (these coordinates are from the deck). Anyone knows what that means for the completion?

atgeirr:
> Anyone knows what that means for the completion?

That completion is completely (hehe) ignored by the simulator.

totto82:
see blattms comments in  OPM/opm-core#1074

jokva:
@totto82 Mind pulling this and try now? We'll ignore completions in inactive cells.

totto82:
It fixed the  `Program threw an exception: Input argument does not correspond to an active cell` problem. But there is still an issue with restart when there is no wells. 
The current status is that the simulator is able to write rst files for cases without wells, but no starting from a report step without wells. Then I still get`Floating point exception (core dumped)` Maybe the return [] fix I suggested not so long ago will fix the problem.

jokva:
How do you run the restart?

totto82:
I use BASIC=2 in RPTRST to run the base case (to generate restart files for every report step) 
In the restart run I use
RESTART
 'path_to_base_case'  restartStep/

It fails for restartStep=1 and 2 (when all wells are shut) 
but works for restartStep>2 when there exists wells

jokva:
@totto82 With the latest patch to the opm-core PR I'm able to restart from step 1.

totto82:
@jokva Hurray. It works. Thanks.

jokva:
Excellent. Please help me test it a bit more, i.e. verify that restarted simulations yield correct(ish) results, and other robustness issues.

jokva:
This PR is largely done, but requires some testing (by @totto82 et al) and rebasing before the actual merge.

Perfect restart probably won't be supported until simulator people figure out a way to also populate controls and completion rates for `WellStateFullyImplicitBlackoil`, but opm-output exposes the data with this patch.

totto82:
@jokva What do you want me to test.

jokva:
That you can restart properly and that restarts yield correct results.

totto82:
@jokva Could you rebase this an fix the merge issues. I am ready to test it.

jokva:
Rebased.

atgeirr:
Looks like this must be rebased. Can you do that @jokva?

jokva:
Rebased.

PR #99, opened 2016-09-27
WIP Output solvent summary

totto82:
The following keywords are implemented
[F,G,W]N[P,I][R,T]
FNPR, WNIR etc.

NOTE 1: 
The total T keywords does not yet work as a fix in ert is needed. 
@joakim-hove Could you add NPT and NIT to the total_vars list in smspec_node.c 

NOTE 2:
The connection keywords are only partly implemented 
(CNFR, CNPT and CNIT) as they are not yet supported from the simulator
side.


joakim-hove:
> The total T keywords does not yet work as a fix in ert is needed. 
> @joakim-hove Could you add NPT and NIT to the total_vars list in smspec_node.c

Yes - will do. Just to be absolutely certain, the keywords you are requesting are: `{FNPT,FNIT,WNPT,WNIT,GNPT,GNIT,CNPT,CNIT}`?

totto82:
Yes.

joakim-hove:
> @joakim-hove Could you add NPT and NIT to the total_vars list in smspec_node.c

Done: https://github.com/Ensembles/ert/pull/1241

joakim-hove:
OK - this looks  perfectly fine; but please add a test - just extend the [test deck](https://github.com/OPM/opm-output/blob/master/tests/summary_deck.DATA) with some solvent related properties - and then actually [check them](https://github.com/OPM/opm-output/blob/master/tests/test_Summary.cpp). 

Since solvent properties is a bit off-the-beaten track I actually consider adding tests specially important.

totto82:
Test is added. 

Note 1: Now output values are not necessarily positive (The CNFR keyword) Please double-check whether my logic here is correct. 
Note 2: I needed to add rates2.set( rt::gas, -20.2 /day) to make the tests pass before I started adding the solvent stuff.

totto82:
@joakim-hove  Thanks for the fix in Ert. It works as expected.

jokva:
Nice work. I do have a suggestion though: we can remove `crate_all` and rather introduce the unary operation `neg` (it used to be there, but it was removed as there used to be no need for it). That way the solvent completion rate function becomes `{ "CNFR", neg( crate< rt::solvent > ) }`

totto82:
Do you mean

``` C++
{ "CNFR", sum ( neg ( injecrate <rt::solvent>), prodcrate <rt::solvent>) }
```

jokva:
I suppose. Isn't the only difference between `crate_all` and `crate` that crate interprets negative input as production, and `crate_all` just returns as-is?

totto82:
Here is a solution without `crate_all`

jokva:
Still seems iffy - what's the flow rate meant to be anyway? The sum of production (negative) and injection (positive  completion rates? What does eclipse output when completions produce more than whatever's injected? Is that positive or negative?

totto82:
The way I understand the CNFR keyword is that it return positive values for producing perforations and negative values for injecting perforations. (i.e the opposite of OPM) 
Since prodcrate and injecrate returns postive values or zero, the minus operator will make sure that 
a non-zero value with the correct sign is returned for both producing and injecting perforations.

jokva:
Right, at the (fairly low cost) of doubling the work. Pretty elegant actually. Mind naming it `sub` instead of `minus`?

totto82:
No problem.

jokva:
Splendid!

jokva:
I'll merge once the tests are green.

PR #101, opened 2016-10-03
Celldata container

joakim-hove:
Simple container for simulators-output interaction. 

opm-core: https://github.com/OPM/opm-core/pull/1081
opm-simulator: https://github.com/OPM/opm-simulators/pull/851
opm-upscaling: https://github.com/OPM/opm-upscaling/pull/207

~~This must be merged first: https://github.com/OPM/opm-common/pull/182~~


atgeirr:
I foresee no major problems with this, but it would probably conflict with OPM/opm-simulators#839, which I have not merged yet since I have not had it tested for the parallel case. Hope to get that out of the way soon though.

pgdr:
1.  Why isn't the underlying data type a map if that's what we're trying to make it look like?
2.  The `Container` postfix doesn't add much to this type, if it's supposed to be a map, maybe call it so?
3.  If it's just a map, why do we need this type, and not just use `std::map`?

joakim-hove:
> 1. Why isn't the underlying data type a map if that's what we're trying to make it look like?
> 2. The Container postfix doesn't add much to this type, if it's supposed to be a map, maybe call it so?
> 3. If it's just a map, why do we need this type, and not just use std::map?

Ahhh - so difficult questions; I have to admit this is done a bit "on a feeling", I foresee some additional metadata, and would like to create a custom container instead of a plain STL structure. If you insist this is premature abstraction I will not really be able to defend myself.

When it comes to `std::vector<T>` versus `std::map<T>` the reason to use `std::vector<T>` is mainly to preserv insert-order.

pgdr:
If I understand correctly, we're going to use insertion order for the actual output?  Is this a feature that Eclipse has that we want to support?

And we're not planning on going over to an underlying map-structure in the future? Or are we possibly adding the insertion-order into `celldata`?

joakim-hove:
> If I understand correctly, we're going to use insertion order for the actual output? Is this a feature that Eclipse has that we want to support?

Yes - that was the intention; I am not necessarily arguing it is very important though.

> And we're not planning on going over to an underlying map-structure in the future? Or are we possibly adding the insertion-order into celldata?

Don't know - as I see that is the advantage of packing the stl stuff in specific container implementation.

jokva:
This feels _very_ much like over-engineering to me, honestly, and I'm not convinced this provides any value over an stl container. It's seems more clunky, doesn't clarify very much and more importantly provides no new behaviour or guarantees (w.r.t. integrity, ordering etc.).

There are some implementation problems as well, but I won't pick on those until what we're doing is clarified.

joakim-hove:
Hehe - this was not exactly all the rage. Personally I like to create custom containers because I feel that increases the semantic content, compared to pure STL containers. For this particular case I foresee the same container to be used for the field properties for both restart files (both the required data like `PRESSURE` and `SWAT` and various auxiallary data like `KRO` ), summary files, init files and rft files - I therefor would like to promote the container beyond a plain STL container. 

Comments about the implementation are welcome.

jokva:
> Personally I like to create custom containers because I feel that increases the semantic content, compared to pure STL containers

Types are nice (I'd be the first to tell you), but I'm still not convinced a thinly-wrapped vector/map actually gives something. Custom STL-derived types shine when they enforce some new integrity or non-obvious property or just a particularly useful interface. _So far_ I'm not convinced we even need this, but I might want to be.

At least for the sake of documentation I'll have a look at details.

opm/output/data/CellDataContainer.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include <algorithm>
+
+#include <opm/output/data/CellDataContainer.hpp>
+#include <opm/output/data/Cells.hpp>
+
+
+namespace Opm {
+
+    CellDataContainer::CellDataContainer( std::vector<data::CellData> data ) {
```
jokva: Pass by value here is good.

opm/output/data/CellDataContainer.cpp
```diff
@@ -0,0 +1,77 @@
...
+        for (const auto& elm : data)
+            this->data.push_back( std::move(elm) );
```
jokva: Proper implementation of the constructor:

`CDC::CDC( std::vector< CD > x ) : data( std::move( x ) ) {}`

opm/output/data/CellDataContainer.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    void CellDataContainer::insert(data::CellData cell_data) {
+        this->data.push_back( std::move(cell_data) );
+    }
+
+
+    const data::CellData& CellDataContainer::getKeyword(const std::string& keyword) const {
+        const auto iter = std::find_if( this->data.begin() , this->data.end() , [&keyword](const data::CellData& cd) { return cd.name == keyword; });
+        if (iter == this->data.end())
+            throw std::invalid_argument("No such keyword in container: " + keyword);
```
jokva: Since checking for existence is removed (it used to be here, right), is it really reasonable to throw if a keyword isn't present?

tests/test_EclipseWriter.cpp
```diff
@@ -285,21 +285,21 @@ BOOST_AUTO_TEST_CASE(EclipseWriterIntegration) {
...
         eclWriter.writeInitAndEgrid( );
         eclWriter.writeInitAndEgrid( eGridProps );
 
         data::Wells wells;
 
         for( int i = first; i < last; ++i ) {
-            std::vector<data::CellData> timesStepProps {
-                {"KRO" , UnitSystem::measure::identity , std::vector<double>(3*3*3 , i)},
-                {"KRG" , UnitSystem::measure::identity , std::vector<double>(3*3*3 , i*10)}};
+            CellDataContainer timesStepProps({
```
jokva: This change is only mandated because `CellDataContainer` doesn't expose `vector`'s rich set of constructors any more, but I suppose that's fine.

Still though, I'd expect a type like this to support init lists if I'd touch upon it in templated code for example.

opm/output/data/CellDataContainer.cpp
joakim-hove: Yes.

opm/output/data/CellDataContainer.cpp
joakim-hove: Well - the `hasKeyword( )` method can be used to check for existence; as long as the return is by  reference I can not see an alternative to throw?

tests/test_EclipseWriter.cpp
joakim-hove: > I'd expect a type like this to support init lists if I'd touch upon it in templated code for example.

Yes I fully agree; would something like this cut it:

``` C++
CellDataContainer::CellDataContainer(std::initializer_list<data::CellData> init_list) 
   : data( init_list )
{ }
```

jokva: Well, at that point you might as well just template-forward all of the underlying constructors. But the implementation itself is correct.

opm/output/data/CellDataContainer.cpp
jokva: That's why `std::map::find` returns an iterator, so no, there's no alternative.

pgdr:
I have no issues in merging this once comments from @jokva is addressed.  However, I really would like to see a different name.  Collection, map, what have you.  If it's an ordered map we're mimicking, maybe we should make it clear that that's what this is.

opm/output/data/CellDataContainer.cpp
blattms: I must admit that I do not understand all this code in detail and I guess that is because one needs all the implementation details for this. Anyway both approaches seem weired to me and might cause quite some surprises for users. Is there some documentation anywhere. If there is not then this is is quite careless.

@jokva is proposing moving from a temporary where I would have assumed that the compiler would have optimized the temporary away and copied directly into data. Furthermore I would assume that moving a vector is quite different from moving its elements. The former results in the vector being invalid, for the latter the vector is valid but its entries are not.. What am I missing?

joakim-hove: The code has been changed to use `std::initializer_list<>` - that was actually the intention all the way; I just did not understand how to it first. I will add documentation.

jokva: > @jokva is proposing moving from a temporary where I would have assumed that the compiler would have optimized the temporary away and copied directly into data. 

It probably will, yes, but it's also crossing module boundaries and I've yet to dig up a proper standards reference for it (I will later tonight probably).

> Furthermore I would assume that moving a vector is quite different from moving its elements. The former results in the vector being invalid, for the latter the vector is valid but its entries are not.. What am I missing?

The `data` argument vector would be well-defined but unspecified, and all its elements would belong to `CDC`. Since `data` is a by-value copy this is perfectly fine.

> The code has been changed to use std::initializer_list<> - that was actually the intention all the way; I just did not understand how to it first. I will add documentation.

You also need the vector overload.

joakim-hove: >  I will add documentation.

Some documentation added to the `.hpp` file. I think it is difficult to strike the balance between useful information and noise when adding documentation.

jokva: > It probably will, yes, but it's also crossing module boundaries and I've yet to dig up a proper standards reference for it (I will later tonight probably).

I've yet to find a good standard reference, but it seems so far that the preferred way is indeed to move out of the passed-by-value copy into the local instance. `CDC::CDC( std::vector x ) : local( std::move( x ) ) {}` should be right, in particular because `std::vector` does not have a pass-by-value constructor (only copy- and move constructor).

joakim-hove: > You also need the vector overload.

Why is that?

jokva: Because the interface is either incomplete or horribly inefficient and clumsy if you _don't_ provide it. The simulator is meant to populate this either directly or construct a vector and then pass that to this class. Except they can't do that if you only support initializer lists because vectors aren't convertible to init lists (nor are they supposed to be - initializer lists are special), meaning you're forcing your clients to create an init list at their site, which is not how they're intended to be used. It also mandates another copying of the elements (into the init list) which is not what you want for runtime-populated containers.

jokva: That was more of a wall than I intended. Some examples:

```

/* simulator: */
std::vector< CD > properties;
properties.populate();
output.writeTimeStep( properties );
```

This might work if we take a vector as now, and then do the conversion to `CDC` later (to no real end, but I digress). However, if we expect a `CDC` in `writeTimeStep`, then `properties` must be it as well. And that limits _their_ usage to the insert method and friends `CDC` provides, meaning you can't use ranges, you can't construct directly from a vector, you can't use non-const iterators etc. Which is probably fine for output, because we want it as a read-only view, but that might not be suitable for the simulator.

Additionally, requiring `CDC` creates a _very_ strong coupling between the simulator and output that isn't actually needed (if you can construct from iterators or vectors).

Having `std::initializer_list` as the _only_ constructor is also a big problem - firstly because it disallows default construction (which you want in this case), but also because it means you can't construct what's essentially a `std::vector` from a `std::vector`, you can only do it by manually copying each element of a `std::initializer_list`.

opm/output/data/CellDataContainer.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Opm {
+
+    CellDataContainer::CellDataContainer( std::initializer_list<data::CellData> init_list )
+        : data( init_list )
+    { }
+
+
+    CellDataContainer::CellDataContainer( std::vector<data::CellData> init_list )
+        : data(std::move(init_list))
```
blattms: Out of curiosity: Has anybody checked what the compiler does here? I would assume that it has to do the copy and hopefully is smart enough to not move. I any case it seem that this will never be better than

```
CellDataContainer::CellDataContainer( const std::vector<data::CellData>& init_list )
        : data(init_list)
```

Which does definitely just copy. And if init_list is a just created temporary then the compiler might be smart enough to do just the construction and skip the copy.

atgeirr: I think that with move semantics, the pass-by-value is better when you actually need to make a copy, precisely because it can move out of a temporary if you pass it. That is a change from traditional C++, where it did not matter and always using a const ref was a simple and good rule.

But no, I have not looked at what a compiler actually does (should be easily testable though).

jokva: The copy may be elided call-site. The move from the temporary is likely avoided, but that probably depends on optimisation levels and more, and can likely vary on a case-by-case basis.

jokva:
General question: why does the simulator concern itself with what's needed to restart from and what's visualisation dumps? Or even init properties?

opm/output/eclipse/EclipseReader.cpp
```diff
@@ -74,40 +74,50 @@ namespace {
...
-        };
+        {
+            const auto apply_pressure = [&]( double x ) {
+                return units.to_si( UnitSystem::measure::pressure, x );
+            };
 
-        const auto apply_temperature = [=]( double x ) {
-            return units.to_si( UnitSystem::measure::temperature, x );
-        };
+            const auto apply_temperature = [=]( double x ) {
```
jokva, line 84: This probably need to capture by reference (well, copying unitsystem isn't the _worst_ thing in the world, but it's not really accomplishing anything).

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -646,12 +625,7 @@ void EclipseWriter::writeTimeStep(int report_step,
        the conversion is done once here - and not closer to the actual
        use-site.
     */
-    convertFromSiTo( cells[ dc::PRESSURE ], units, UnitSystem::measure::pressure );
-    if ( cells.has( dc::TEMP ))
-        convertFromSiTo( cells[ dc::TEMP ],
-                         units,
-                         UnitSystem::measure::temperature );
-
+    cells.convertFromSI( units );
```
jokva, line 625: Personally I don't like this kind of API because it's confusing and it's somewhat flaky (converting from SI twice is possible unless you keep internal state to make this method a no-op on successive calls). Can we lift the method and keep `cells` more oblivious to units?

opm/output/eclipse/EclipseWriter.hpp
```diff
@@ -92,14 +92,13 @@ class EclipseWriter {
...
                         data::Wells,
-                        const CellDataContainer& simProps = {},
                         bool write_float = true);
 
     EclipseWriter( const EclipseWriter& ) = delete;
     ~EclipseWriter();
 
 private:
-    void writeINITFile( const CellDataContainer& simProps, const NNC& nnc) const;
+    void writeINITFile( data::Solution simProps, const NNC& nnc) const;
```
jokva: Can these be removed from the header while we're at it?

alfbr:
> General question: why does the simulator concern itself with what's needed to restart from and what's visualisation dumps? Or even init properties?

I believe you need to put the question in a more specific context before it makes sense. A simulator obviously cannot care about anything. From a theoretical point of view, you can ask if differences between these data need to be exposed in the simulator code at all. I believe the answer is yes. For instance, what is needed for a restart depends on the simulator implementation itself. Same goes for initial properties actually (assuming here that is what you meant, rather than the slightly different concept of an INIT file). From the more pragmatic angle it may more be a question of whether it makes sense for that complexity to be exposed in the simulator code. Personally, I think the answer there is yes, even it the theoretical part is ignored by locking in on one simulator implementation (as we have more or less done so far). You may have asked for something entirely different though.

joakim-hove:
> General question: why does the simulator concern itself with what's needed to restart from and what's visualisation dumps? Or even init properties?

The simulator does not care - but the output layer does; and hence the simulator needs to tell.

joakim-hove:
Ehhh - wrong button.

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -597,7 +578,7 @@ void EclipseWriter::writeEGRIDFile( const NNC& nnc ) const {
 }
 
 
-void EclipseWriter::writeInitAndEgrid(const std::vector<data::CellData>& simProps, const NNC& nnc) {
+void EclipseWriter::writeInitAndEgrid(data::Solution simProps, const NNC& nnc) {
```
jokva, line 577: This can probably be a reference, right?

joakim-hove: I really think so yes.

opm/output/eclipse/EclipseWriter.hpp
joakim-hove: OK - they have been added to the private implementation.

opm/output/eclipse/EclipseWriter.cpp
joakim-hove: Have added a internal bool flag to the container which keeps track of whether the data is in SI units or not.

pgdr:
jenkins build this with downstreams opm-core=1081 opm-simulators=851 please

pgdr:
This needs a corresponding PR in upscaling before it's ready:

```
.../upscaling/SteadyStateUpscalerImplicit_impl.hpp:269:54:
    error: 'Opm::data::Solution::key' has not been declared
        solution.insert( data::Solution::key::PRESSURE, ecl_press );
```

opm/output/eclipse/EclipseWriter.cpp
```diff
@@ -737,24 +708,25 @@ void EclipseWriter::writeTimeStep(int report_step,
...
-                        auto ecl_data = grid.compressedVector( prop.data );
-                        convertFromSiTo( ecl_data,
-                                         units,
-                                         prop.dim );
+                for (const auto& prop : cells) {
+                    if (prop.second.target == data::TargetType::RESTART_AUXILLARY) {
+                        auto ecl_data = grid.compressedVector( prop.second.data );
+                        //convertFromSiTo( ecl_data,
+                        //                 units,
+                        // prop.dim );
```
pgdr: why is this here?  Maybe better to add a written comment if there's an intention behind it.  Else remove.

tests/test_CellDataContainer.cpp
```diff
@@ -0,0 +1,86 @@
...
+#if HAVE_DYNAMIC_BOOST_TEST
+#define BOOST_TEST_DYN_LINK
+#endif
+
+#define BOOST_TEST_MODULE CellDataContainer
+#include <boost/test/unit_test.hpp>
+
+#include <vector>
+
+#include <opm/output/data/CellDataContainer.hpp>
```
pgdr: This file should be deleted, no?

pgdr: tests/test_CellDataContainer.cpp that is

pgdr (approved):
This PR should be merged,  HOWEVER ...
1. upscaling needs updating
2. test_CellDataContainer needs to be deleted

opm/output/eclipse/EclipseWriter.cpp
joakim-hove: Removed.

tests/test_CellDataContainer.cpp
joakim-hove: Yes - done.

joakim-hove:
OK - all comments addressed, and a new PR in opm-upscaling: https://github.com/OPM/opm-upscaling/pull/207 has been created. Will merge the whole shebang when https://github.com/OPM/opm-upscaling/pull/207 has been approved.

pgdr:
jenkins build this with downstreams opm-core=1081 opm-simulators=851 opm-upscaling=207 please

PR #103, opened 2016-10-10
update packaging for new dependency tree

akva2:


joakim-hove:
This looks correct - and as long as building binary packages is not part of the CI there should be no point in waiting for that to complete.

PR #105, opened 2016-10-10
adapt to the rename of the unit system header in opm-parser

andlaus:
see OPM/opm-parser#936 for details.


andlaus:
jenkins build this opm-parser=936 opm-core=1085 please

PR #106, opened 2016-10-11
Region production

joakim-hove:
OK - this should actually be ready now. With this in place we get support for the summary keywords: ROIP, ROIPL, ROIPV, ROPT, RGPT, RWPT.

Observe that this requires: https://github.com/Ensembles/ert/pull/1285 to pass all tests.

### DIFF
--- a/opm/output/eclipse/RegionCache.cpp
+++ b/opm/output/eclipse/RegionCache.cpp
@@ -18,6 +18,10 @@
  */
 
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/CompletionSet.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Completion.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Well.hpp>
 #include <opm/parser/eclipse/EclipseState/Eclipse3DProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/GridProperty.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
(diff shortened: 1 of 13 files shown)


jokva:
Starting to look good!

Just a general observation: most std::containers have `empty` in addition to `size`. Other than that it's generally preferred to use `iostream` rather than printf, mostly because mixing the two isn't all that nice and printf's less flexible and type safe, and other opm modules already use iostream.

joakim-hove:
> Other than that it's generally preferred to use iostream rather than printf

Yes - the `printf( )` was just debugging; but my oh my quick and dirty formatted output with `iostream` is painful.

jokva:
> Yes - the printf( ) was just debugging; but my oh my quick and dirty formatted output with iostream is painful.

amen brother, it's dreadful. It has plenty of nice features, but is slow and clumsy for quick one-offs.

atgeirr:
> quick and dirty formatted output with iostream is painful.

Oh yes. For me, C printf is also painful... The only language I have ever found it easy to format output in was perl, unfortunately!

jokva:
Perl is love, perl is life. Perl's great, python isn't too bad, I think Haskell is nice too and I suppose the lisps have some niceness to them (at least if you pimp it a bit with macros).

Regardless, iostreams are clunky but they do have some nice features. Rare you need or get to use them, but still.

opm/output/eclipse/RegionCache.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <opm/parser/eclipse/EclipseState/Eclipse3DProperties.hpp>
+#include <opm/parser/eclipse/EclipseState/Grid/GridProperty.hpp>
+#include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
+
+#include <opm/output/eclipse/RegionCache.hpp>
+
+namespace Opm {
+namespace out {
+
+    RegionCache::RegionCache(const EclipseState& state, const EclipseGrid& grid) {
```
jokva, line 34: My only wonder w.r.t. this is the name cache - why does it matter that it's a cache? Describing functionality and all.

opm/output/eclipse/RegionCache.hpp
```diff
@@ -32,10 +32,14 @@ namespace out {
         RegionCache() = default;
         RegionCache(const EclipseState& state, const EclipseGrid& grid);
         const std::vector<size_t>& cells( int region_id ) const;
+        const std::vector<std::pair<std::string,size_t>>& completions( int region_id ) const;
 
     private:
-        std::vector<size_t> empty;
+        std::vector<size_t> cells_empty;
```
jokva, line 38: These should probably be static members to save some space.

tests/test_Summary.cpp
```diff
@@ -182,7 +182,7 @@ struct setup {
         ta( ERT::TestArea("test_summary") ),
         solution( make_solution( es.getInputGrid() ) )
     {
-        solution = make_solution( *es.getInputGrid());
+        solution = make_solution( es.getInputGrid());
```
jokva, line 211: Not really related to this PR, but anyone know why solution isn't just in the init list?

tests/test_Summary.cpp
```diff
@@ -632,3 +633,35 @@ BOOST_AUTO_TEST_CASE(region_vars) {
...
+    BOOST_CHECK( ecl_sum_has_general_var( resp , "ROPR:1"));
+    BOOST_CHECK_CLOSE(ecl_sum_get_general_var( resp , 1 , "ROPR:1" ) ,
+                      ecl_sum_get_general_var( resp , 1 , "COPR:W_1:1") +
+                      ecl_sum_get_general_var( resp , 1 , "COPR:W_2:2") +
+                      ecl_sum_get_general_var( resp , 1 , "COPR:W_3:3"), 1e-5);
+
+
+    BOOST_CHECK( ecl_sum_has_general_var( resp , "RGPT:1"));
+    BOOST_CHECK( ecl_sum_has_general_var( resp , "CGPT:W_1:1"));
+    /*BOOST_CHECK_CLOSE(ecl_sum_get_general_var( resp , 2 , "RGPT:1" ) ,
```
jokva: Commented out.

jokva (changes requested):

jokva:
Superseded by https://github.com/OPM/opm-output/pull/122

PR #108, opened 2016-10-13
Avoid init warning due to missing init for struct member.

atgeirr:
It would be good to have a test that new flag has the intended effect too, but this PR only silences the warning.


joakim-hove:
This will be in conflict with #101 which I am really keen to merge. Closing.

atgeirr:
I see your point, just aiming for a clean compile before I started checking out and reviewing code again (such as #101 and its consequences...)

I guess also OPM/opm-simulators#858 may be in conflict with the stuff you have in flight? That one also contains compile error fixes, so it would be good to get it in (this PR on the other hand is fully obsolete I guess?)

PR #109, opened 2016-10-14
The great shared_ptr purge

jokva:
Downstream of https://github.com/OPM/opm-parser/pull/939

In addition to plain fixes from interface changes in opm-parser, this moves the CornerpointChopper.hpp file to opm-upscaling, as it never belonged in opm-output.


pgdr (approved):
Very good!  Merge when Jenkins says go!

PR #110, opened 2016-10-19
fix build on clang++ 3.8

andlaus:
this is the clang version which is shipped with ubuntu 16.10. it says:

```
default initialization of an object of const type 'const
std::unordered_map<int, std::vector<size_t> >' (aka 'const
unordered_map<int, vector<unsigned long> >') without a user-provided
default constructor
```

I don't know if that error is correct or not, but initializing the
object uppon construction should not hurt...


jokva:
I was pretty sure it would be default-initialised, but not so much anymore.

andlaus:
as mentioned: it may equally well be a clang bug...

andlaus:
hm: I can reproduce the test failure on my machine, but I've got not idea why it happens. On the other hand, this patch does not seem necessary any more because the latest master version now compiles with clang on my Ubuntu 16.10 machine. closing.

PR #112, opened 2016-10-20
Using constructor in default argument creation.

joakim-hove:


jokva:
Why? Genuinely curious if it makes a difference.

joakim-hove:
> Why? Genuinely curious if it makes a difference.

Well - on Travis it actually does; don't ask me why.

joakim-hove:
I'll self merge this now - as it resurrects the travis build chain.

jokva:
I think it might be the defaulted Solution constructor in conjuction with using map. The default one should go, I'll submit a patch.

atgeirr:
In general, inheriting from classes not designed to be base classes (such as map) carries a significant risk of surprises, although I am not sure that is what bit you this time.

PR #114, opened 2016-10-23
Will printf the number of keywords when error.

joakim-hove:
Minor improvement in debug output. Should aid in debugging the current regression test failure.

### DIFF
--- a/opm/test_util/EclFilesComparator.cpp
+++ b/opm/test_util/EclFilesComparator.cpp
@@ -19,6 +19,9 @@
 #include <opm/test_util/EclFilesComparator.hpp>
 #include <opm/common/ErrorMacros.hpp>
 
+#include <stdio.h>
+
+#include <set>
 #include <iostream>
 #include <iomanip>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


pgdr (approved):

akva2:
jenkins build this with downstreams opm-data=136 please

akva2:
the failure is due to opm-parser and its shared libraries. but i cannot fathom why this bites now and not earlier. for some reason, the opm-output binaries gets their rpath stripped while other modules keeps em. i'm currently investigating

joakim-hove:
> the failure is due to opm-parser and its shared libraries.

OK - that was indeed surprising. Would it be possible to run the tests on jenkins as `ctest --output-on-failure`\- then it would be easier to debug.

akva2:
what would that add that's not already there? all output is accessible from the job logs, that adds nothing afaik.

joakim-hove:
> what would that add that's not already there? all output is accessible from the job logs, that adds nothing afaik.

Maybe not much, but if you compare these two:

[travis](https://travis-ci.org/OPM/opm-output/builds/169979625) and [jenkins](http://80.240.129.131:8080/job/opm-output-PR-builder/15/consoleFull) I would say the former is more usable.

akva2:
i respectfully completely and utterly 100% disagree. travis gives absolutely no results from tests, jenkins give 100%  you just have to look at the _test output_ http://80.240.129.131:8080/job/opm-output-PR-builder/15/testReport/, not the console output.

atgeirr:
> you just have to look at the test output http://80.240.129.131:8080/job/opm-output-PR-builder/15/testReport/, not the console output.

That... is useful! Wish I had known earlier (have to click on "Test results" in the first screen to get to it).

akva2:
jenkins build this with downstreams opm-data=136 opm-common=193 please

akva2:
builds is passed anything that should be influenced by changes and all tests passed. this, along with the two other pr's can be merged. i have to run to a meeting now so i suggest one of you merge when the build is finished. if not i will handle in a few hours.

PR #116, opened 2016-10-24
data::Well::Completion is vector, not map

jokva:
The data exchanged in data::Well::Completions is now a vector, more
closely matching the disk format and the order exposed by the simulator.
Zero'd wells and complations are still written for shut wells, or wells
the simulator does not provide any information for; however, all
non-closed wells (according to the schedule) will be restored. The
completions are written and restored in the same order as CompletionSet
from parser.


jokva:
@atgeirr

atgeirr:
Thanks, I'll make a followup in opm-simulators.

Edit: and in opm-core, of course.

atgeirr:
Followups have been created, and successfully tested locally. Waiting for Jenkins and @totto82 approval for merging.

PR #118, opened 2016-10-25
Fix warnings

atgeirr:
This fixes some warnings encountered with clang. It is done on top of #116, so it should not be merged as-is, I will rebase it after #116 is merged.

The new content here are the commits ~~7a425c7~~ 6a61f30 and ~~07d3bc7~~ d70cb0d.


atgeirr:
Rebased now.

jokva (approved):

PR #121, opened 2016-10-26
Add phases to summary_deck.DATA

anrdnet:
summary_deck.DATA used for tests doesn't contain any phases.

WCONHIST then sets control mode to ORAT, which doesn't make sense if we have no phases.

This is a problem for OPM/opm-parser#953.


joakim-hove:
Thank you - this is obviously right?

joakim-hove:
Thank you - this is obviously right?

joakim-hove:
Will merge when I have access to a proper workstation.

pgdr:
> Will merge when I have access to a proper workstation.

Not if I do it first.

PR #122, opened 2016-10-28
Model2 missing out kws

jokva:
Support for GGORH, GWCTH, RGIP, FOIP and FGIP.

PR #124, opened 2016-11-01
Pass NUM == 1 in when determining units.

joakim-hove:


atgeirr:
This seems to allow Norne to continue past init on my system (running in debug mode, so it has not gotten very far yet, but it has done a complete report step).

However, I get a test failure in test_Summary:

```
/Users/atgeirr/opm-build-release/opm-output/tests/test_Summary.cpp:688: error in "region_production": difference{50%} between ecl_sum_get_general_var( resp , 2 , "RGPT:1" ){300.39999389648438} and ecl_sum_get_general_var( resp , 2 , "CGPT:W_1:1") + ecl_sum_get_general_var( resp , 2 , "CGPT:W_2:2") + ecl_sum_get_general_var( resp , 2 , "CGPT:W_3:3"){600.79998779296875} exceeds 1.0000000000000001e-05%
```

This happens both for debug and release versions. Do you get this, too?

jokva:
Update your ert.

joakim-hove:
The requirement to update ert was in the original PR description: #106 - but the PR which was eventually merged was #122 - there the message had fallen out. Sorry about that.

atgeirr:
Yep, seems to work now. Thanks for the hint!

PR #130, opened 2016-11-04
Check that completion corresponds to active cell.

joakim-hove:


blattms:
Is the explanation what this patch does in OPM/opm-simulator#892 really correct?

In that case I would double check the solution before merging this.

blattms:
Please holdoff merging until we find the real cause of the problem described in OPM/opm-simulators#892

joakim-hove:
> Please holdoff merging until we find the real cause of the problem described in OPM/opm-simulators#892

My take on this is:
1. I think this is a good improvement/fix, which I would like to merge anyway.
2. I agree that this fix does not address MPI specifically, so my hunch is still that Norne is currently failing on master - also for serial runs. If it turns out this an MPI issue, then I am slightly tempted to say it is an unrelated MPI bug.

**Context:** In the `SUMMARY`configuration in the input file you can ask for results per completion. There is no checking in that configuration that the completed cell is indeed active, and the exception reported in  OPM/opm-simulators#892 comes about when the summary code tries to look up an inactive cell. Since the cells queried for in completion output is pure user input - it should be checked, as as is implemented in this patch.
1. I would like to merge this patch anyway.
2. If there is a difference in active cells between the serial and the MPI version - which seems to be the case, that is in my opinion bug in the MPI code which should be fixed _before_ this patch is merged.

blattms:
I am not exactly sure what your patch does. It definitely removes the
check whether the index is active.

Master works serial (both on my laptop and server). In parallel it works on
my laptop (intel core i7), but throws on my server (Intel Xeon E5).

joakim-hove:
> I am not exactly sure what your patch does. 

The interesting part of the patch is [this change](https://github.com/OPM/opm-output/pull/130/commits/428ca7400100aa41cfb273b326eec04613aff1ec#diff-71e13cd4acdd582de8151509e6afa2faR714)

> It definitely removes the check whether the index is active.

That is actually not true; with the new code the inactive/active is checked at an earlier stage than previously.

joakim-hove:
@blattms: as the real problem of the parallell output has now been identified - I would like to merge this (Which will again mask the parallell output problem - is that OK by you?)

blattms:
There would not be thrown any exception anymore with this PR, because of the introduced continue. I prefer getting hit in the face by a bug.

joakim-hove:
> There would not be thrown any exception anymore with this PR, because of the introduced continue. I prefer getting hit in the face by a bug.

Yes - I am known the share that view; in this particular case the `throw` accidentally actually pointed us to a real problem. However the safe guard with the `continue` is still required for situations like this:

``` txt
SUMMARY

-- We request output of of total water production from the completion in cell (2,3,4)
CWPT
 'W_1' 2 3 4  /
/
```

The summary write code must detect the situation, and disable the output, in the case where the cell `(2,3,4)` is inactive - that is what is ensured with the current PR.

joakim-hove:
OK - I guess https://github.com/OPM/opm-simulators/pull/900 solves the real issue. Since this PR solves _another_ issue - I will merge it now.

PR #131, opened 2016-11-05
Remove unused files

joakim-hove:
The writeECLData has been suggested moved to opm-upscaling; the EclipseUnits.hpp is just plain unused.

This should only be merged after: https://github.com/OPM/opm-upscaling/pull/209


atgeirr:
I realized now that this not only moves the old writeECLData function, but also means the old reordering polymer sim no longer can call it (missed that point initially). That makes me very sceptical, as it provides a simulation capability not really matched by the fully implicit simulators (flow variants).

I therefore ask that it is kept around until we have replaced it. It is not quite clear to me why it should be removed, other than the fact that its functionality overlaps with the EclipseWriter?

joakim-hove:
> I realized now that this not only moves the old writeECLData function, but also means the old reordering polymer sim no longer can call it (missed that point initially). 

I also missed that point initially - have create a PR: https://github.com/OPM/opm-simulators/pull/898

> That makes me very sceptical, as it provides a simulation capability not really matched by the fully implicit simulators (flow variants).

The simulator itself is not removed - it is "only" the ability to write binary restart files which is removed; and if that is deemed important the simulator should be updated to use the proper EclipseWriter.

> It is not quite clear to me why it should be removed, other than the fact that its functionality overlaps with the EclipseWriter?

Well - having old unmaintained cruft around _significantly_ slows down development.

atgeirr:
> the simulator should be updated to use the proper EclipseWriter

I guess I should take a look at this. Can I have a few days to evaluate (no time today) before you merge this?

joakim-hove:
> Can I have a few days to evaluate (no time today) before you merge this?

Of course!

jokva:
I suppose a couple of days have passed no, @atgeirr? I'd like to close this.

atgeirr:
> I suppose a couple of days have passed no, @atgeirr? I'd like to close this.

It has, and in the time since I have restored the ability to run the old sim cases, but I have not yet come to the point of checking that the output is ok (removed vtk output got in the way). Is there a reason this has become urgent?

If not, I hope you'll give me a bit more time: it is still near the top of my todo-list.

jokva:
Not urgent, but it's been three weeks, and these things are easily forgettable.

atgeirr:
This and the downstream PR can be merged now (unless you want another test run first).

PR #132, opened 2016-11-08
Summary keywords

jokva:
Some more summary keywords.


pgdr (approved):
Sweet, it really shows that this implementation of `Summary` pays off!

Merge at will!

PR #134, opened 2016-11-10
Summary constructor overloads; minor bugfix

jokva:
Two consecutive merges that did not account for interface changes.


opm/output/eclipse/Summary.cpp
```diff
@@ -700,6 +700,17 @@ class Summary::keyword_handlers {
 };
 
 Summary::Summary( const EclipseState& st,
+                  const SummaryConfig& sum ) :
+    Summary( st, sum, st.getInputGrid() )
```
joakim-hove, line 704: I don't want this - there is so much confusion with the different grids, and adding a convenience constructor which is _generally_ the wrong choice (i.e. using the input grid), will in my opinion only add confusion.

PR #140, opened 2016-11-17
FOE support

jokva:
Field efficiency: (OIP(initial) - OIP(now)) / OIP(initial). The initial
OIP is cached, and FOIP is reused for OIP(now). Adds the
operator-(double,quantity) to make the formula obvious from the FOE
function.

@atgeirr @babrodtk I'd appreciate some help on the simulator side of this. Somewhere around https://github.com/OPM/opm-simulators/blob/master/opm/autodiff/SimulatorBase_impl.hpp#L190 you'd need to call `setInitial` with OIP passed.

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -134,6 +134,10 @@ struct quantity {
     }
 };
 
+quantity operator-( double lhs, const quantity& rhs ) {
+    return { lhs - rhs.value, rhs.unit };
+}
+
 /*
  * All functions must have the same parameters, so they're gathered in a struct
  * and functions use whatever information they care about.
@@ -150,6 +154,7 @@ struct fn_args {
     const data::Solution& state;
     const out::RegionCache& regionCache;
     const EclipseGrid& grid;
+    double initial_oip;
 };
 
 /* Since there are several enums in opm scattered about more-or-less
@@ -411,7 +416,10 @@ quantity foip( const fn_args& args ) {
              measure::volume };
 }
 
-
+quantity foe( const fn_args& args ) {
+    const quantity val = { foip( args ).value, measure::identity };
+    return (args.initial_oip - val) / args.initial_oip;
+}
 
 template< typename F, typename G >
 auto mul( F f, G g ) -> bin_op< F, G, std::multiplies< quantity > >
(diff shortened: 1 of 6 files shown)


pgdr (approved):
This looks good and ready to be merged.  But perhaps we should let Atgeirr and André chime in before merging.

atgeirr:
I'll ask @babrodtk to look at this, I need to concentrate on the Frankenstein merge.

jokva:
It won't really do any new output (save writing zeros) until `setInitial` is called, so it's sort-of harmless to merge, since simulator support can be added independently.

pgdr:
I agree it's safe to merge.

However, this is an "API change (or addendum)" and therefore; perhaps the users of the API would have opinions.

jokva:
Absolutely.

jokva:
I've updated this slightly and removed the `setInitial` method. Instead, you need  to pass "OIP" to `writeInitAndEgrid`, which has now been renamed `writeInitial`. There might be more properties later that requires some knowledge of the initial state, and it's likely that they too will need information through `writeInitial`.

This patch will fail on travis now - when the simulator is ready to adopt this, please merge (even if you're not actually passing OIP). Norne and friends don't use the `FOE` keyword, so it won't break any regression test through keyword support alone.

opm/output/eclipse/Summary.cpp
```diff
@@ -801,6 +819,13 @@ void Summary::add_timestep( int report_step,
     this->prev_time_elapsed = secs_elapsed;
 }
 
+void Summary::set_initial( const data::Solution& sol ) {
```
joakim-hove, line 822: I don't know if we will need more of these initial properties; but at least for the second case I think we should use a map-like structure to internalize these initial values.

joakim-hove:
Will you make a `s/writeEgridandInit/writeInitial/` PR in opm-simulators?

opm/output/eclipse/Summary.cpp
jokva: That's my plan if/when that shows up, as this solution obviously doesn't scale (but is delightfully simple).

jokva:
Yeah, I can do that.

opm/output/eclipse/Summary.cpp
joakim-hove: OK - I am also a proponent of jit bridge burning.

PR #148, opened 2016-12-06
Remove writeVtkData().

atgeirr:
This removes the vtk-writing functions from opm-output.

The important version of the function was commented out here anyway, since it requires the UnstructuredGrid. It has been moved downstream to opm-simulators and resurrected there.


atgeirr:
jenkins build this please

atgeirr:
I forgot to write: this must be merged simultaneously with OPM/opm-simulators#968

joakim-hove:
I would assume  OPM/opm-simulators#968 could be merged independently first?

atgeirr:
> I would assume OPM/opm-simulators#968 could be merged independently first?

Perhaps, I'm not sure having the same symbol in two different libs won't cause trouble. In any case, both are Jenkins-approved now, so feel free to merge!

joakim-hove:
> In any case, both are Jenkins-approved now.

Well - the Jenkins approval of this PR is not that interesting, as long as downstream is not included. But the jenkins approval of: OPM/opm-simulators#968 proves the point about independence.

Anyway. Merging

PR #153, opened 2017-01-09
Fix units for RS and RVwhen loading from restart.

joakim-hove:
Downstream fix: https://github.com/OPM/opm-simulators/pull/1008

Will probably also require updates to the reference data.

### DIFF
--- a/opm/output/eclipse/EclipseReader.cpp
+++ b/opm/output/eclipse/EclipseReader.cpp
@@ -105,13 +105,17 @@ namespace {
 
             /* optional keywords */
             if( ecl_file_has_kw( file, "RS" ) ) {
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this with downstreams opm-simulators=1008 please

atgeirr:
Jenkins did not seem to run this, but the result in the downstream run seem to confirm that the restart reference data must be updated. I do not quite understand all of it though: why did comparing RS output between Flow and ECL reference data succeed before? Or have we only been comparing against Flow references? @akva2?

joakim-hove:
> Or have we only been comparing against Flow references? @akva2?

We are only comparing `flow` &lrarr; `flow`

joakim-hove:
jenkins build this with downstreams opm-simulators=1008 please

PR #155, opened 2017-01-11
Passing correct unit system ID when writing.

joakim-hove:
There is an integer in the `INTEHEAD` keyword in the INIT and RESTART files which value indicates the unit system used in the rest of the file. We have up until now hardcoded the numerical value corresponding to METRIC units, so for anything different from METRIC units there has been an internal inconsistency in the files. With this PR we write the correct id value.

The reference data must be updated. I am sorry about this high rate of reference data changes. Am currently going through the code writing restart files, and small issues arise. 

NB: must be merged together with: https://github.com/Ensembles/ert/pull/1399

PR #156, opened 2017-01-16
Solution data::solution has SI units in Summary.

joakim-hove:
The solution data which was passed to the summary for calculation of e.g. `BPR` was converted from SI units twice. The current PR implements a temporary fix to facilitate update of reference data.

This is is the third time in a short period which has required updates to the reference data, I assume there will be more. Common to these three updates is the following:

1. The error has been gross, wrong units, and a naive comparison with the Eclipse reference solution would have revealed it immediately.
2. The error has been in keywords/results no one has really been interested in, and the results have been printed to output in a sort of whatever-hang along mode.

To reduce the effort and noise related to updating the reference data I wish we could change the testing to:

1. More active use of the eclipse data as reference data.
2. More selective approach to what is actually compared in the regression tests, i.e. we should consciously say that now e.g. 'BPR' is correctly calculated - from now on I want regression testing on this keyword.

### DIFF
--- a/opm/output/eclipse/EclipseWriter.cpp
+++ b/opm/output/eclipse/EclipseWriter.cpp
@@ -684,7 +684,7 @@ void EclipseWriter::writeTimeStep(int report_step,
     const auto& es = this->impl->es;
     const auto& grid = this->impl->grid;
     const auto& units = es.getUnits();
-
+    const data::Solution cells_si = cells;
     const auto& restart = es.cfg().restart();
 
 
@@ -818,7 +818,7 @@ void EclipseWriter::writeTimeStep(int report_step,
                                       this->impl->es,
                                       this->impl->regionCache,
                                       wells ,
-                                      cells );
+                                      cells_si );
     this->impl->summary.write();
  }
 


alfbr:
> This is is the third time in a short period which has required updates to the reference data, I assume there will be more. Common to these three updates is the following:

Yes, and one way to reduce that noise is to collect output of new vectors in fewer pull requests.

> More active use of the eclipse data as reference data.

This is currently not in the pipe-line. What is currently planned is pretty much tailored to the current "make one pull request for every new output vector" approach from developers, namely
-all vectors that are currently present in output should be properly quality assured
-check that vectors present before an update match the same vectors after the update, if they don't, then manual quality assurance of the deviation is needed
-new vectors will need to be quality assured manually (this we may need to put a process in place for)
This should address the current fall-outs in the simplest possible manner.

Using Eclipse reference solutions as the gold standard may be tempting, but it is important to realise that we are not reimplementing Eclipse, only the Eclipse file formats. We do not expect close to identical results with Eclipse. As our code mature, we may even deviate strongly from Eclipse results (e.g., by replacing the two-point scheme with something that handles grid orientation). Hence, we need to have our own tests against our own quality assured reference data regardless. These reference data will need to be updated every time new vectors are added to the output data regardless. The effort to automate this is under way, please be patient.

joakim-hove:
> Yes, and one way to reduce that noise is to collect output of new vectors in fewer pull requests.

That is obviously right, the reason I have done this in multiple small PR's is to reduce the risk of errors in the process.

> -all vectors that are currently present in output should be properly quality assured

Well - that is just not the case; and due to the nature of these file formats that will remain difficult.

joakim-hove:
jenkins build this with downstream please

alfbr:
> Well - that is just not the case; and due to the nature of these file formats that will remain difficult.

I am not sure I follow you. Assuming that we quality assure all current reference data, which we are in the process of doing, how does the file format prevent a subsequent incremental approach?

joakim-hove:
@akva2: Should I use e.g. https://github.com/OPM/opm-simulators/pull/1028 to update the reference data, or will you do it one last time?

As I wrote [here](https://github.com/OPM/opm-simulators/pull/1027#issuecomment-272811442) I *really* think this should be a parameterised jenkins job - any hope for that?

notspiff:
I can of course do it but you using the published scripts would be a nice acid test.

I will look into the jenkins job but not this week, as i am at the geilo winter school, network is shitty, only have my laptop etc.

joakim-hove:
> I can of course do it but you using the published scripts would be a nice acid test.

OK mr *spiff* - I will give it a shot.

akva2:
Sigh wrong device/login. Me ^

joakim-hove:
> I am not sure I follow you. Assuming that we quality assure all current reference data, which we are in the process of doing, how does the file format prevent a subsequent incremental approach?

Well - I am not saying the file format prevents an incremental approach, but it is at least cumbersome. Take for instance the keyword `INTEHEAD` which is part of the restart and init files. That keyword should have a number of integer values, like e.g. the cartesian dimensions of the grid, simulation time and so on. The values in the `INTEHEAD` kyeword come in three categories:

1. Values we handle correctly.
2. Values which are documented, but which we have not yet come around to support.
3. Values which are not documented - which we typically just set to zero.

`INTEHEAD[2]` is an integer saying which unit system which used in the file, previously that was randomly hardcoded to the metric value, now we are gradually improving and as part of #155 we set `INTEHEAD[2]` to the correct integer value. This is certainly an incremental improvement, but since the regression testing is based on whole-keyword testing we must update the reference data. I would much have preferred a testing framework which allowed me to say:  Now `INTEHEAD[2]` is promoted from a random hardcoded value to something which I care about -  please included it in the regression testing." This will be a recurring theme  for e.g. the `XWEL` and `IWEL` keywords in the restart files.

joakim-hove:

alfbr:
> `INTEHEAD[2]` is an integer saying which unit system which used in the file

Yes, the auxiliary (or should I say meta-data) information in the files are a bit particular. Still, the examples you provided seems to fit well into the incremental approach:
-If these vectors are already present in the reference solution, we should check that no other vectors are altered significantly (I guess we can put some effort into automating this step too if warranted)
-If these are new vectors, we need to check that none of the old vectors are altered significantly (this will hopefully be automated soon)
-In both cases we need to have a process in place to ensure that the new or significantly different vectors are quality assured

It sounds like you are against checking if the same vectors we output yesterday are still output today, and rather have a fixed list of output vectors we check, while not monitoring any change in output files besides these. This is easily achievable with the current test-framework, but I believe it is a step back. My apologies if I misunderstood.

joakim-hove:
jenkins build this with downstreams opm-data=158 please

akva2:
jenkins build this with downstreams opm-data=159 please

PR #159, opened 2017-01-23
Testing automation

akva2:
I need a PR to stick the noise into while implementing automation of data updates. Just ignore this..


akva2:
jenkins build this with downstreams opm-common=225 opm-simulators=1038 update_data please

PR #161, opened 2017-02-02
Add option to not throw on each failure in summary regression tests

akva2:
Useful for performing further analysis of the failure.

### DIFF
--- a/opm/test_util/summaryComparator.hpp
+++ b/opm/test_util/summaryComparator.hpp
@@ -26,6 +26,17 @@
 #include <algorithm>
 #include <string>
 
+
+// helper macro to handle error throws or not
+#define HANDLE_ERROR(type, message) \
+  { \
+    if (throwOnError) \
+      OPM_THROW(type, message); \
+    else \
+      std::cerr << message << std::endl; \
+  }
+
+
 //! \brief Prototyping struct, encapsuling the stringlist libraries.
 struct stringlist_struct;
 typedef struct stringlist_struct stringlist_type;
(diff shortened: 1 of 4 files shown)


opm/test_util/summaryRegressionTest.cpp
```diff
@@ -60,7 +62,7 @@ void RegressionTest::getRegressionTest(){
                 if (isRestartFile && keywordString.substr(3,1)=="T"){
                     break;
                 }
-                checkForKeyword(timeVec1, timeVec2, keyword);
+                throwAtEnd = !checkForKeyword(timeVec1, timeVec2, keyword);
```
joakim-hove: Shouldn't this be something like `throwAtEnd |= !checkForKeyword( );` to carry a `true` value through the keyword list?

opm/test_util/summaryComparator.hpp
```diff
@@ -26,6 +26,17 @@
 #include <algorithm>
 #include <string>
 
+
+// helper macro to handle error throws or not
+#define THROW_OR_NOT(type, message) \
```
joakim-hove: I would suggest `#define ERROR` - the current name seems to indicate that the `THROW_OR_NOT` macro can answer the question: "Should I throw?".

joakim-hove:
This is a welcome addition - personally I would like to see the default behavior to be `throwOnError == false`.

opm/test_util/summaryComparator.hpp
```diff
@@ -26,6 +26,17 @@
...
+  { \
+    if (throwOnError) \
+      OPM_THROW(type, message); \
+    else \
+      std::cout << message << std::endl; \
```
joakim-hove: `std::cerr` ?

opm/test_util/summaryRegressionTest.cpp
akva2: nice catch.

opm/test_util/summaryComparator.hpp
akva2: which is precisely what this macro does, no ? (through the if()). anyways, i'll rename..

opm/test_util/summaryComparator.hpp
akva2: indeed. naughty attempt at aiding my scripts, hoped nobody would notice ;)

akva2:
as for throwing by default, i respectfully disagree. it has the potential to spam the terminal a bunch, and i strongly believe that should be opt-in.

akva2:
rebased with fixes.

PR #162, opened 2017-02-06
fixed: sanity check size of data array

akva2:
This fixes the norne through flow/flow_mpi/flow_ebos after recent refactorings.

I am not sure if this is just a missing sanity check, a change in semantics or a broken deck. I have thus also opened https://github.com/OPM/opm-simulators/pull/1046 which makes sure the caller does not push the empty vectors in the first place.


akva2:
@alfbr as requested.

joakim-hove:
Can you explain the situation where this happens? My initial feeling is that this looks more like an internal error/bug which should lead to an exception.

akva2:
at time step 1, after substep 0, the simulator pushes empty GIPG/OIP/et al fields into the output container. thus, the check for the field existence passes, but the data vector is empty.

akva2:
typo. it's after the first substep which is labeled substep 0.

joakim-hove:
> at time step 1, after substep 1, the simulator pushes empty GIPG/OIP/et al fields into the output container. thus, the check for the field existence passes, but the data vector is empty.

OK - but this is a simulator bug, we can merge the alternative code:

```C++
f (sim_value.size() < cells.size())
     throw std::runtime_error("Passing empty error to summary write");
```
but I see no reason to paper over simulator bugs in this way.

akva2:
sure, you do as you see fit, i have offered solutions in both repos as well as played with open cards. a throw certainly does not solve the regression however.

joakim-hove:
I'll call.

akva2:
hold for a better error message please

akva2:
there we go, sorry for delay, skype intervened
'Error: Program threw an exception: Wrongly sized data array passed to output for keyword GIPG, size=0, expected=3504.'

joakim-hove:
Thank you; absolutely last nit-pick: `if (sim_value.size() != cells.size()) { ... `.

akva2:
i was faster with my rewrite ;) (already there)

akva2:
not sure how to address the test failure, to my uneducated eye it seems the input data is to blame ?

akva2:
right. after closer inspection it seems that the throw is wrong. we are passed the global data, but cells only holds a subset (the region). i guess we can eat the slight overhead of a max_element here?

joakim-hove:
> not sure how to address the test failure, to my uneducated eye it seems the input data is to blame ?

Ehhh - no; took me some to disentangle this too. The `cells` property is a a set of indices which should be summed over; the test should be: `if (sim_value.size() != args.grid.getNumActive()) { ... `

akva2:
numActive is 999 in the test, not 1000.

joakim-hove:
> numActive is 999 in the test, not 1000.

OK - *that* is input-data failure; I will make a PR for that.

akva2:
surpassed by #163

PR #163, opened 2017-02-06
Fix testdata size

joakim-hove:
Continuation of: #162

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -351,6 +351,13 @@ quantity region_sum( const fn_args& args , const std::string& keyword , UnitSyst
         return { 0.0, unit };
 
     const std::vector<double>& sim_value = args.state.data( keyword );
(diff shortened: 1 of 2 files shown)


opm/output/eclipse/Summary.cpp
```diff
@@ -351,6 +351,13 @@ quantity region_sum( const fn_args& args , const std::string& keyword , UnitSyst
         return { 0.0, unit };
 
     const std::vector<double>& sim_value = args.state.data( keyword );
+    if (sim_value.size() != args.grid.getNumActive()) {
+      std::stringstream str;
+      str << "Wrongly sized data array passed to output for keyword "
+          << keyword << ", size=" << sim_value.size() << ", expected=" << cells.size() << ".";
```
akva2: cells.size -> args.grid.getNumActive()

joakim-hove: Correction force pushed.

PR #167, opened 2017-02-15
Document init

joakim-hove:
+ Write NTG. Completely untested.

### DIFF
--- a/opm/output/eclipse/EclipseIO.hpp
+++ b/opm/output/eclipse/EclipseIO.hpp
@@ -53,19 +53,54 @@ public:
 
 
 
-    /// Write the static eclipse data (grid, PVT curves, etc) to disk, and set
-    /// up additional initial properties
-    ///
-    /// - simProps contains a list of properties which must be
-    ///   calculated by the simulator, e.g. the transmissibility
-    ///   vectors TRANX, TRANY and TRANZ.
-    ///
-    /// - The NNC argument is distributed between the EGRID and INIT
-    ///   files.
-    ///
-    /// If you want FOE in the SUMMARY section, you must pass the initial
-    /// oil-in-place "OIP" key in Solution
-    void writeInitial( data::Solution = data::Solution(), const NNC& = NNC());
+/**    \brief Output static properties in EGRID and INIT file.
+  *
+  *    Write the static eclipse data (grid, PVT curves, etc) to disk,
+  *    and set up additional initial properties. There are no specific
+  *    keywords to configure the output to the INIT files, it seems
+  *    like the algorithm is: "All available keywords are written to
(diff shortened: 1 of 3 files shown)


andlaus:
I guess this also needs some simulator side support to be useful?

joakim-hove:
Well. - if the simulator wants to output something funky the simulator must calculate that funky property and pass it in the `simProps` object. But API wise this is ready and has been for quite some time.

Of course I have to fix the compile issue, and there will be updated to the reference data (I think) - but then this can be merged.

andlaus:
> this is ready and has been for quite some time.

Does this mean that NTG will be written by flow* after this PR?

joakim-hove:
Well - iff NTG is part of the deck: yes.

15. feb. 2017 18:46 skrev "Andreas Lauser" <notifications@github.com>:

andlaus:
cool :+1:

opm/output/eclipse/EclipseIO.hpp
```diff
@@ -53,19 +53,53 @@ class EclipseIO {
...
-    ///   calculated by the simulator, e.g. the transmissibility
-    ///   vectors TRANX, TRANY and TRANZ.
-    ///
-    /// - The NNC argument is distributed between the EGRID and INIT
-    ///   files.
-    ///
-    /// If you want FOE in the SUMMARY section, you must pass the initial
-    /// oil-in-place "OIP" key in Solution
-    void writeInitial( data::Solution = data::Solution(), const NNC& = NNC());
+    /*
```
atgeirr: Just a minor thing: this changes the doc string from being a doxygen doc string to just a regular comment. Was that intended?

joakim-hove: Not really - I'll revert that part of the commit.

alfbr:
That was fast, thanks!

>  iff NTG is part of the deck: yes.

Actually, I believe NTG is written out regardless. I can double check tomorrow and report back. According to the documentation, it is defaulted to 1 for every cell after the GRID section.

alfbr:
I can confirm that NTG is always written out to the INIT file. If it is not set in the input files, then all values are defaulted to 1.

joakim-hove:
> I can confirm that NTG is always written out to the INIT file. If it is not set in the input files, then all values are defaulted to 1.

OK good - it should now be written unconditionally to the INIT file.

alfbr:
Great, then I have no further comments.

joakim-hove:
Some fiddling in parser required first: https://github.com/OPM/opm-parser/pull/1043

joakim-hove:
jenkins build this with downstreams opm-parser=1043 please

pgdr (approved):
Looks good!  Merge!

PR #168, opened 2017-02-16
added: option to not throw on error in restart/init regression test

akva2:
useful for doing a more thorough analysis of the failure.


joakim-hove:
Will this do `exit(1)` if there are errors and `throwOnError == false`?

akva2:
nope, good catch.

akva2:
rebased.

PR #169, opened 2017-02-17
Added well load test to restart testing.

joakim-hove:
Added call to load wells as test.


joakim-hove:
@akva2: The second commit here is in your part of the woods.

akva2:
Just suffered a horrible os crash so right now i am working on restoring from backup.. will look into it asap but probably monday..

akva2:
looks fine.

joakim-hove:
> looks fine.

I'll take that as a review :smile:

PR #170, opened 2017-02-22
Added ability to save restart fields in double.

joakim-hove:
Observe that just the scaling between SI units and output units is enough to loose bitwise equality.

### DIFF
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -386,7 +386,8 @@ void EclipseIO::writeTimeStep(int report_step,
                               bool  isSubstep,
                               double secs_elapsed,
                               data::Solution cells,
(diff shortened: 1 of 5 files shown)


atgeirr:
> Observe that just the scaling between SI units and output units is enough to loose bitwise equality.

While not bitwise equal I expect that it will be close enough that the solvers (linear and nonlinear) will never make different choices (in original and restarted run) because of this. Thanks for providing the ability!

joakim-hove:
For people investigating restart I encourage you to look at the `RestartIO::load()` and `RestartIO::save()` functions, using those functions you can completely bypass deck configuration of restart times.

atgeirr:
Just tested this -- works beautifully!

joakim-hove:
> Just tested this -- works beautifully!

Good - I am glad to hear!

PR #171, opened 2017-02-24
[WIP] Do not throw for field sizes different from number of cells.

atgeirr:
Work in progress, do not merge (until after discussion at least).

This simple change allows the user to read (and write, that is already possible) arbitrary-sized fields through the opm-output restart load/save functionality. It does mean that CellData no longer must contain cell data, so if adopted we should also do some renaming. An alternative (and perhaps better) solution could be to add new function argument(s) for communicating arbitrary fields.


joakim-hove:
Closing this - alternative here: https://github.com/OPM/opm-output/pull/172

PR #172, opened 2017-02-27
Extra data in restart files.

joakim-hove:


joakim-hove:
Will require some impedance matching in opm-simulators; will @atgeirr take that part?

atgeirr:
> will @atgeirr take that part?

I will.

opm/output/eclipse/RestartValue.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+    struct RestartValue {
+
+        data::Solution solution;
+        data::Wells wells;
+        std::map<std::string,std::vector<double>> extra = {};
+
+
+        RestartValue(data::Solution sol, data::Wells wells, std::map<std::string, std::vector<double>> extra) :
+            solution(sol),
```
atgeirr: Should use std::move(sol) etc. here to take advantage of move semantics, I believe (which was why you did not use const ref arguments I suppose).

joakim-hove: Yes - fixed.

joakim-hove:
Hmmmm - it works on my machine :angry:

joakim-hove:
jenkins build this please

joakim-hove:
@akva2: What was the trick to use `valgrind` as test command?

opm/output/eclipse/RestartValue.hpp
```diff
@@ -0,0 +1,58 @@
...
+    */
+
+    struct RestartValue {
+
+        data::Solution solution;
+        data::Wells wells;
+        std::map<std::string,std::vector<double>> extra = {};
+
+
+        RestartValue(data::Solution sol, data::Wells wells, std::map<std::string, std::vector<double>> extra) :
```
atgeirr: Shadowing generates annoying warnings on my system. Could we change the argument names to wells_arg and extra_arg for example?

joakim-hove: Sure

joakim-hove: Do the tests pass on your machine?

akva2:
ctest -T memcheck

opm/output/eclipse/RestartValue.hpp
atgeirr: Yes they do.

joakim-hove:
> ctest -T memcheck

Thank you.

CMakeLists.txt
```diff
@@ -91,3 +91,7 @@ macro (install_hook)
 endmacro (install_hook)
 
 include (OpmLibMain)
+
+add_test( NAME valgrind-restart
+          COMMAND valgrind ${PROJECT_BINARY_DIR}/bin/test_Restart
+          WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/tests)
```
andlaus: I guess you need to add `CONDITION ${VALGRIND_FOUND}` here because this will only work if valgrind is installed...

CMakeLists.txt
```diff
@@ -91,3 +91,7 @@ macro (install_hook)
 endmacro (install_hook)
 
 include (OpmLibMain)
+
+add_test( NAME valgrind-restart
+          COMMAND /usr/bin/valgrind ${PROJECT_BINARY_DIR}/bin/test_Restart
```
andlaus: is hardcoding the path to valgrind a good idea? what happens if I compiled it myself and installed it into a custom location?

atgeirr:
The valgrind test fails for me (do not have valgrind installed).

joakim-hove:
The (hardcoded!) valgrind stuff is purely WIP; I am just trying to understand why the test fails on Travis.

atgeirr:
The jenkins run from opm-simulators was successful, are there still things you want to do here? (Other than possibly squashing or cleanup.)

joakim-hove:
I am uncomfortable merging before the Travis failure is understood/resolved.

atgeirr:
> I am uncomfortable merging before the Travis failure is understood/resolved.

I get that. I'll see if I can provoke a failure on my system.

opm/output/eclipse/RestartIO.cpp
```diff
@@ -509,8 +541,10 @@ void save(const std::string& filename,
           data::Wells wells,
           const EclipseState& es,
           const EclipseGrid& grid,
+          std::map<std::string, std::vector<double>> extra_data,
```
atgeirr: I looked at this and found no errors. There are a few minor things I would change, though, that I did not notice before. The `wells` and `extra_data` arguments should be passed as const references (unlike `cells` they are not modified inside this function). Also, I'd put the `reserved_keys` inside the checking scope below. Not that it should matter, but could there be a problem with multithreaded calls to this function, because of the nontrivial static variable (perhaps someone with more knowledge about this can enlighten me)? I think removing the `static` optimization would not harm performance in any case since this is after all a function that writes to disk.

I cannot reproduce, but on Travis the error is triggered in the function that uses the default parameter. Perhaps removing that default parameter could shed some light on it?

joakim-hove:
Thank you for the suggestions and comments - I will try to look at this
afternoon.

2. mars 2017 09:03 skrev "Atgeirr Flø Rasmussen" <notifications@github.com>:

opm/output/eclipse/RestartIO.cpp
joakim-hove: >The wells and extra_data arguments should be passed as const references 

The reason these are passed by value is to support asyncronous output, where calling scope can manipulate these variables immediately after the call is issued.

atgeirr: Ok, I think I see your point, but in the asynchronous case we store all variables inside the functor that makes the write call so I think it should not actually matter here though.

atgeirr:
I see that Travis tested ok (in opm-output) now. Do you know what caused the previous problems?

joakim-hove:
> Do you know what caused the previous problems?

To be honest no - but I think the failure happened in a code path mixing C++ exceptions and C code; should in my opinion only have resulted in a resource leak and not a full blown error, but anyway rewritten now,

atgeirr:
> code path mixing C++ exceptions and C code

That sounds like it could easily blow up... I'm curious about where we have such a mix though?

joakim-hove:
> I'm curious about where we have such a mix though?

It was in the `RestartIO::save( )` function which first opened a ert handle to an output file, and then could throw a C++ exception when the size of the cells data was wrong. Anyway - fixed now.

PR #179, opened 2017-04-07
Made extra data support also float data

babrodtk:
This is required for OPM/opm-simulators#1084 in the updated version which supports restarting from old restart files (which do not have SOMAX, and the other hysteresis parameters).


atgeirr:

joakim-hove:
I can merge this alright - but the PR description and the actual change do not match up. The current PR will *not* help with:

>  [...]  from old restart files (which do not have SOMAX, and the other hysteresis parameters).

However; as the the title says it will help with loading restart data in `float` format; but have we suddenly started writing `float` - or do you actually intend to load restart files with `SOMAX` which have been generated by Eclipse - and therefor use `float`?

So - all in all I do not really understand?

joakim-hove:
When I finally understood what this was all about i must say I really did not like it. I will make an alternative based on optional keys to the restart loader.

PR #180, opened 2017-04-11
Restart optional

joakim-hove:


opm/output/eclipse/RestartIO.cpp
```diff
@@ -249,8 +255,9 @@ RestartValue load( const std::string& filename,
 
         if (ecl_file_view_has_kw( file_view , key.c_str())) {
             const ecl_kw_type * ecl_kw = ecl_file_view_iget_named_kw( file_view , key.c_str() , 0 );
-            std::vector<double> data = double_vector( ecl_kw );
-            rst_value.extra[ key ] = std::move(data);
+            const double * data_ptr = ecl_kw_get_double_ptr( ecl_kw );
```
atgeirr, line 258: Won't this fail if the file contains the keyword but it is stored as floats?

joakim-hove: Yes it will; but the extra keywords are written in `double`, and now with optional loading of restart  keywords there should be no need to mix the two classes of keywords.

atgeirr: There are a few things I did not see:
 - I could not find where the optional keywords are set to be written with `double`? And how are they separated from the non-optional ones at write time (the optionality is only passed in at read time).
 - Should we not read both `float` and `double` interchangably in any case?

joakim-hove: > I could not find where the optional keywords are set to be written with double?

Well - that is not configurable; extra keywords are always written in double.

> And how are they separated from the non-optional ones at write time 

Well, at write time everyting which is passed to `EclipseIO::writeRestartFile( )` is written to disk. I can not see that it makes sense to tell the writer - at write-time - that in a subsequent invocation down the line reading of this keyword will be optional?!

> Should we not read both float and double interchangably in any case?

We could of course decide to do that - the reason I was so keen on reverting the `float -> double` conversion was that I felt it was a crutch which enabled an inelegant "wrong" solution to the problem.

atgeirr: Ok I think I get your point: extra keywords are read as double, but we no longer use this path to read the now-optional keywords so it is ok. For regular (non-extra) keywords (optional or not) we *do* read float or double interchangeably. Correct?

joakim-hove: > For regular (non-extra) keywords (optional or not) we do read float or double interchangeably. Correct?

That is correct.

joakim-hove:
jenkins build this with downstreams opm-simulators=1136 please

joakim-hove:
See: https://github.com/OPM/opm-simulators/pull/1136#issuecomment-293406672

PR #181, opened 2017-04-12
Silence warnings

atgeirr:
Minor fixes that should have absolutely no effect other than removing compile-time warnings. I hope this can get in when green, @andlaus, @joakim-hove?

### DIFF
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -95,13 +95,6 @@ void writeKeyword( ERT::FortIO& fortio ,
 
 
 
-/// Convert OPM phase usage to ERT bitmask
-inline int ertPhaseMask( const Phases& phase ) {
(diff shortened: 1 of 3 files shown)


atgeirr:

atgeirr:
Ping?

andlaus:
from my POV, this can be also pushed into the release branch once it is in master.

akva2:
just in case: nobody touches the branches before i say they are unfrozen please.

andlaus:
> just in case: nobody touches the branches before i say they are unfrozen please.

ok, fine. will you do the cherry picking then? how do you want to be notified? Anyway, I'd also like to get OPM/opm-simulators#1146 into the release if it gets merged into master in time.

akva2:
i will release the branches after i have built the rc1 packages, then maintainers (or more perhaps release manager) own the branches again. point is i cannot package a moving target. i'm ~ halfway there.

PR #182, opened 2017-04-19
Fixes some problems with option parsing.

blattms:
Namely, that m requires an argument and not n as indicated to
getopt. This caused the help message to be printed whenever users
did not want exceptions to be thrown.

Additionally we fix the error message for options that need an argument
but do not get one.


akva2:
thanks. i must have been disengaged brain when i added that option.

PR #185, opened 2017-05-09
HCPV weight FPR

totto82:
Tested on rel 0 model 2.

Reference data may need to be updated if FPR is compared in the tests.


opm/output/eclipse/Summary.cpp
```diff
@@ -853,6 +861,13 @@ void Summary::add_timestep( int report_step,
 
     const size_t timestep = report_step;
     const auto& schedule = es.getSchedule();
+    const std::vector<double>& porv_all = es.get3DProperties().getDoubleGridProperty("PORV").getData();
+    const size_t  num = this->grid.getNumActive();
+    std::vector<double> porv (num,0.0);
+    for (size_t active_index = 0; active_index < num ; ++active_index) {
+        const size_t global_index = this->grid.getGlobalIndex(active_index);
```
joakim-hove: This is available as the `compressedCopy( )` property of the PORV property.

joakim-hove: A bit excessive to make this compressed copy for every single timestep - a `porv` property should be added as a private member to the `Summary` class - and initialized in the constructor.

joakim-hove:
> Tested on rel 0 model 2.

Well - that does not provide a safety net for the developers coming after you; please add a test e.g. [here](https://github.com/OPM/opm-output/blob/master/tests/test_Summary.cpp#L526)

joakim-hove:
> Reference data may need to be updated if FPR is compared in the tests.

Yes - the reference data must be updated.

joakim-hove:
@andlaus  / @atgeirr : the `TRANX` fauilure in the Norne init test seems a bit suspicious, and probably not related this PR?

totto82:
> Well - that does not provide a safety net for the developers coming after you; please add a test e.g. here

FPR is already tested. Since porosity and swat is constant in the domain the FPR value is unchanged by this PR for this test. Do you want me to change the porosity in summary_deck.DATA to be no-constant and change the reference solution accordingly.

totto82:
I have compared FPR on the 2D_H__ test for multisegmentedwells,  with this PR the FPR values are approx the same as Eclipse. (without it the deviation is significant)

joakim-hove:
> FPR is already tested

Well; you have changed the implementation; the existing test accidentally still works, that is not the same as "FPR is already tested" - on the contrary in my opinion. The Right&trade; way to do this would in my opinion have been to:

1. Create a test with a non-uniform PORO / SWAT and verify that the new test breaks with the existing `FPR` implementation.
2. Modify the `FPR` implementation until all tests pass.
3. Commit and make PR - and *forget everything about `FPR` calculations*.

> Do you want me to change the porosity in summary_deck.DATA to be no-constant and change the reference solution accordingly.

Yes. I reccomend making a new test and deck, to avoid breaking what is already there.

totto82:
Now with tests

joakim-hove:
> Now with tests

Very good - thank you.

This PR in itself is ready to be merged - @totto82: please push the green button when the updated reference data has been prepared.

totto82:
jenkins build this update_data please

joakim-hove:
> jenkins build this update_data please

I *think* the `update_data` feature is only available from opm-simulators; i.e. you must make a dummy PR there to initiate the process.

akva2:
sorry - it's on my list to get this available everywhere but it hasn't popped off the stack yet.

akva2:
lol. memory like a gold fish. i *HAVE* added the update-data triggers everywhere before. just remember to include downstreams when using it from non-simulators repos.

PR #191, opened 2017-06-04
Ability to pass misc values to summary writer.

joakim-hove:


atgeirr:
Having looked at (only) the single change in EclipseIO.hpp I think that looks fine at least! I'll try to implement TCPU using this as a base.

opm/output/eclipse/Summary.cpp
```diff
@@ -736,13 +733,30 @@ static const std::unordered_map< std::string, ofun > funs = {
...
+  {"TIMESTEP" , UnitSystem::measure::time },
+  {"TCPUTS"   , UnitSystem::measure::time },
+  {"TCPUDAY"  , UnitSystem::measure::time },
+  {"STEPTYPE" , UnitSystem::measure::identity },
+  {"TELAPLIN" , UnitSystem::measure::time }
+};
+  
+  
+
+  inline std::vector< const Well* > find_wells( const Schedule& schedule,UnitSystem::measure:: },
```
atgeirr: There is some weirdness here, and it won't compile. Removing the UnitSystem... part makes it compile.

joakim-hove: Hmmm - that was an unfortunate fat-fingering; fixed.

atgeirr:
This is finished now in your view, right @joakim-hove?

joakim-hove:
I consider this as finished yes - if you as 'customer' agree I guess we are all set.

atgeirr:
I agree, we have something that can be merged. We could wait until it has been tested though, as discussed in the other PR.

PR #193, opened 2017-06-08
Fixes for warnings from static code analysis

akva2:
Question: Is suppression using inline comments acceptable? (last commit).

### DIFF
--- a/opm/output/data/Solution.hpp
+++ b/opm/output/data/Solution.hpp
@@ -34,7 +34,7 @@ class Solution : public std::map< std::string, data::CellData > {
 
     public:
         Solution() = default;
-        Solution( bool si );
+        explicit Solution( bool si );
         using Base::map;
         using Base::insert;
 
--- a/opm/output/data/Wells.hpp
+++ b/opm/output/data/Wells.hpp
@@ -43,6 +43,7 @@ namespace Opm {
          */
 
         public:
+            Rates() = default;
             enum class opt : uint32_t {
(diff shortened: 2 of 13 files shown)


joakim-hove:
So to answer question: I would prefer if it were fixed without comment based suppression, and I guess at least in this case that can be easily achieved?

joakim-hove:
Not clear to me why it is failing - has a real problem been unearthed?

joakim-hove:
Address the comments you agree with and merge when the test failure has been resolved.

akva2:
failure needs investigation, i did not encounter it locally.

akva2:
jenkins build this please

akva2:
i cannot reproduce the problem on either of my 2 machines, or 3 vms, or jenkins :/. valgrind is silent, so is gdb, with or without symbols.

are you sure travis is working? i see you are pulling packages from lots of different places. in particular pulling boost backports built with standard toolchain, and then pulling in a toolchain backport is rather live fast, die young kind of behavior.

joakim-hove:
> i cannot reproduce the problem on either of my 2 machines, or 3 vms, or jenkins :/. valgrind is silent, so is gdb, with or without symbols.

OK - thank you for thorough investigations.

> are you sure travis is working?

I guess this question is not on the ultimate list of questions we are really certain about - I thought so.

> and then pulling in a toolchain backport is rather live fast, die young kind of behavior.

Good point.

PR #195, opened 2017-06-19
Added method to verify report steps.

joakim-hove:
Followup to: https://github.com/OPM/opm-simulators/pull/1222

Check that the reports steps from two summary cases agree.


joakim-hove:
Hmmm - this little test revealed that the summary writer does not add the restart header, must fix that first.

joakim-hove:

PR #198, opened 2017-07-19
Allow simulator to overwrite initial OIP value.

blattms:
Up to now we assumed that if there is interest in the
initial OIP value (e.g. to calculate FOE) then oip has
to be presented to EclipseWriter during the call of
writeInitial. For the downstream simulators OIP is
not available at this stage. This commit gives the simulator
the possibility to overwrite/reset the values later and
allows the current implementations to let output calculate
and output FOE.

### DIFF
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -373,6 +373,10 @@ void EclipseIO::writeInitial( data::Solution simProps, const NNC& nnc) {
 }
 
 
+void  EclipseIO::overwriteInitialOIP( const data::Solution& simProps )
+{
+    this->impl->summary.set_initial( simProps );
+}
 
 // implementation of the writeTimeStep method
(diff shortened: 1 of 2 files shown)


joakim-hove:
I argue that the downstream PR: https://github.com/OPM/opm-simulators/pull/1243 is not entirely correct, if that PR is changed to calculate the `OIP` values prior to calling the `EclipseIO::writeInitial( )` method this PR should not be required any longer.

If you indeed wish to merge https://github.com/OPM/opm-simulators/pull/1243 in it's current form - please feel free to also merge this PR.

blattms:
The problem is that we need the full grid and transmissibilities for writeInitial, but the OIP is only available on the distributed grid. Using the writeTimeStep functionality is just a convenient way to gather it for opm-output.

Of course it would be better to be able to call writeInitial on the distributed data, but that seems a lot more work (at least for legacy) for little gain.

joakim-hove:
Ok - please merge it for now.

PR #199, opened 2017-07-25
Groups of groups summary data.

joakim-hove:
When evaluating a group variable, like e.g. GOPR we need to traverse the
group tree all the way down to the leaf nodes, i.e. wells.


joakim-hove:
jenkins build this with downstreams opm-parser=1111 please

PR #202, opened 2017-09-25
Cmakify sibling search

blattms:
Same procedure as in OPM/opm-material#247


joakim-hove:
OK - if this is The right way to go&trade; - I will of course not interfere. I guess the approach has already been vetted by the build master @akva2?

PR #208, opened 2017-10-09
WIP: changed sequence of KWs to comply with eclipses standard

stefoss23:
This is a very small part of the project of making OPM restart files eclipse readable. 
One requirement is that nothing uneclipse-like should be added after the STARTSOL keyword.
The proposed change here changes the keyword sequence.


andlaus:
I'm sympathetic to the goal and your patch seems to be quite okay to me (be aware that I don't call the shots in this module), but if eclipse is *this* picky about keyword ordering et al. keeping the restart files compatible seems to be a major long term pain in the butt from the maintenance perspective!?

stefoss23:
Actually the project is currently on hold due to the complexity of the task.

PR #213, opened 2017-11-14
Add group keywords

lars-petter-hauge:


lars-petter-hauge:
Added GGPRS and GGPRF to output.
 
Included `rt::solution_gas` as a new rate to be populated by output from the simulator.

Could this be a valid option? or perhaps an alternative name of "solution_gas"? @atgeirr @joakim-hove

atgeirr:
I am not sure what solution_gas is, is it gas dissolved in oil (gas-oil-ratio) or solvent gas (which we use for modeling co2 eor processes)?

lars-petter-hauge:
gas initially dissolved in oil yes (indeed, part of gas-oil-ratio, Rs), not solvent.

atgeirr:
So this would be part of the gas rate? I.e. `Rates::gas == Rates::solvent_gas + (gas that was free gas in the reservoir)`?

lars-petter-hauge:
yes.

joakim-hove:
As main responsible for pushing the big green in opm-output I am ready to merge this; but I must admit it is not entirely clear to me how good the impedance match between `rt::solution_gas` and the simulator is. If @atgeirr is happy we merge this - otherwise we must find an alternative.

atgeirr:
I'm ok with it, with a few questions:
 - Is solution_gas the term most commonly used for this (and not, say, dissolved_gas)? (In documentation or papers etc.)
 - What about the "symmetric" case of vaporized oil?

lars-petter-hauge:
* Solution gas drive and dissolved gas drive are used interchangeably.

* Given a parameter `vaporized_oil` it would perhaps make more sense with `dissolved_gas`. You will find "Solution Oil" defined in the eclipse manual, although in the literature "Gas Condensate" is a much more used terminology.

atgeirr:
I very much prefer "vaporized oil" over "gas condensate" in the code, as that is what we have called it in formal and informal contexts. So I think I prefer "dissolved_gas" to "solution_gas" here, but will not veto you.

Next question: I assume this needs to be wired from the simulator side to be useful, so when is that needed...?

lars-petter-hauge:
Yes, I agree. I mentioned gas condensate rather as a response to "common terms", instead of what the parameter name should be.

I've changed the name from solution_gas to dissolved_gas, and I'll use vaporized_oil for the next keywords.

Furthermore the output from the simulator, which I frankly am not familiar with yet, will need to populate Rates with dissolved_gas. Not sure if I'm following you, what do you mean by "when is that needed"?. As in what situation?

atgeirr:
> what do you mean by "when is that needed"?. As in what situation?

No, more like "today", "next week" or "before the summer"...

joakim-hove:
This is still marked ask 'WIP' - but apart from  that I am ready to merge.

lars-petter-hauge:
Last push was a quick fix to allign parameters of Rates.

The consequence of merging this pull request will be that the keyword "GGPRF" will be calculated as Gas Production Total - dissolved gas. The value of `dissolved_gas,` as for the other `Rates`, is by default 0.0. In the case that this pr is merged, and the simulator does not populate the Rates with dissolved_gas, the output from GGPRF will always be the same as GGPR (Gas Production Total) and output from GGPRS will always be 0.0. This should be fixed before the next release I suppose.

lars-petter-hauge:
I removed the WIP on this. 

@atgeirr alright if we go ahead and merge?

atgeirr:
Sure, no problem!

PR #215, opened 2017-11-16
Add Support for Writing Saturation Function Tables to INIT File

bska:
This change-set extends the INIT file Table writing to also output the saturation function tables in normalised form.  We support generating the INIT file's SGFN, SOFN, and SWFN tables&mdash;including the derivative information&mdash;for both two and three phases from both families of saturation function keywords (S{G,W}OF and S{G,W}FN + SOF{2,3}).  We do not yet support generating the appropriate tables from input keyword SLGOF.

The main part of this change-set is commit ~~35eac9f~~ 07aad01 which adds the table normalisation and padding for all combinations of supported saturation function keywords and a collection of unit tests to verify that we produce the tables we expect.  This commit builds on commit ~~3a3738c~~ 81d64c7 which introduces a new helper class to simplify indexing into the linearised and padded "TAB" vector entries and enables using standard algorithms such as `std::copy()` and `std::transform()` to assign the appropriate column values rather than using explicit looping.  This commit also adds a helper function,
```C++
Opm::DifferentiateOutputTable::calcSlopes()
```
that is useful as a means of centralising the calculation of derivatives of the piecewise linear interpolants that are implicitly specified through the normalised saturation function tables.  Derivative/slope information must accompany the primary output columns according to the file format protocol.

As part of creating the unit tests I have elected to remove the member function
```C++
void Opm::Tables::fwrite(ERT::FortIO& fortio) const
```
in favour of providing equivalent functionality through a new free function,
```C++
void Opm::fwrite(const Tables& tables, ERT::FortIO& fortio)
```
that is implemented in terms of acquiring read-only references to the internal representations of the "TABDIMS" and "TAB" vectors (see commit ~~346d977~~ c3cb948) and forming `ERT::EclKW<>` instances based on those data values.

Do note that as this change-set explicitly modifies the "INIT" file, we will break any regression test that compares the contents of that file before and after this change.  We should therefore only install this if we're comfortable updating the relevant baseline files.

Finally, a few notes about the implementation:

1. We create scoped helper functions to handle each saturation function table in the appropriate context, for instance
    ```C++
    std::vector<double>
    SatFunc::Oil::TwoPhase::fromSGOF(const std::size_t          numRows,
                                     const Opm::TableContainer& sgof);
    ```
    to create the two-phase "SOFN" vector entries from information stored in tables entered with the `SGOF` keyword.

2. Each of these scoped helper functions are, in turn, implemented in terms of a helper template function
    ```C++
    template <class BuildDependent>
    std::vector<double>
    createSatfuncTable(const std::size_t numTab,
                       const std::size_t numRows,
                       const std::size_t numDep,
                       BuildDependent&&  buildDeps)
    ```
    that implements a simple loop over saturation function regions and adds appropriate derivative information to each normalised output table.  The callable entity `buildDeps`&mdash;not necessarily a good name&mdash;is responsible for assigning independent variables and the table-required dependent variables (e.g., relative permeability for water and oil-water capillary pressure) for a single table/saturation region and is invoked on each iteration of `createSatfuncTable<>()`.

3. Creating the three-phase "SOFN" table from family one (SGOF, SWOF) keywords is slightly involved.  We need to merge the tables based on common (and unique) values of the oil saturation which tends to expand the number of rows needed to represent the table.  This expansion in turns means that missing relative permeability values must be defined through piecewise linear interpolation in the "KrOX" column of the pertinent input table.  We defer the task of creating a unique sequence of merge oil saturation values to the standard library routine `std::set_union()` and then only need to create a wrapper class to facilitate oil saturation look-up and relative permeability value either in terms of a saturation node index or a saturation value.


joakim-hove:
This was a mighty piece of work; will probably need some to review (i.e. understand ...) it. Epic commit messages :-)

opm/output/eclipse/LinearisedOutputTable.hpp
```diff
@@ -0,0 +1,168 @@
...
+        /// Destructive access to internal data buffer.
+        ///
+        /// Mostly to support outputting all table data to external storage.
+        ///
+        /// \return \code std::move() \endcode of the internal data buffer.
+        std::vector<double> getDataDestructively();
+
+    private:
+        /// Internal buffer for tabular data.
+        std::vector<double> data_;
```
joakim-hove: I'm sorry to bother you with this - but please please; those trailing underscores hurt my eyes - there is no tradition for using them in the rest of opm-output.

bska: I can probably name it `m_*` if that's better, but I can't keep the existing name as there can't be two class-scope identifiers called `tabdims`.

bska:
> This was a mighty piece of work

Thanks.  I've been building it on and off for the past month or so.  Many iterations...

> will probably need some to review (i.e. understand ...)

There's no rush.  Feel free to ask questions about it.

joakim-hove:
> Feel free to ask questions about it.

I guess I will - mainly for my own interest and curiosity; are you on Texas time or Gaustad time?

joakim-hove:
> [... ] keywords is slightly involved. 

I recognize an understatement when I see one.

bska:
> are you on Texas time or Gaustad time?

Gaustad.

>> [... ] keywords is slightly involved.
> 
> I recognize an understatement when I see one.

In that case, take comfort in the knowledge that it would be much more complicated if `std::set_union()` did not exist.

bska:
I have rebased this change-set onto current master and addressed a few issues that presented during review.  I now consider this to be the more-or-less final version of the work and would like to submit it to final approval.

joakim-hove:
jenkins build this with downstreams update_data please

joakim-hove:
jenkins build this with downstreams opm-data=279 please

PR #216, opened 2017-11-17
Added possiblity: writing integer fields to INIT files.

stefoss23:
An extra argument of type std::map has been added to EclipseIO::writeInitial:
```C++
void EclipseIO::writeInitial( data::Solution simProps, std::map<std::string, std::vector<int> > map, const NNC& nnc)
```
All entries in map will be written to INIT file in ecl keyword format.

### DIFF
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -192,7 +192,7 @@ inline std::string uppercase( std::string x ) {
 class EclipseIO::Impl {
     public:
     Impl( const EclipseState&, EclipseGrid, const Schedule&, const SummaryConfig& );
-        void writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > map, const NNC& nnc) const;
+        void writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > int_data, const NNC& nnc) const;
         void writeEGRIDFile( const NNC& nnc ) const;
 
         const EclipseState& es;
@@ -221,7 +221,7 @@ EclipseIO::Impl::Impl( const EclipseState& eclipseState,
 
 
 
-void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > map, const NNC& nnc) const {
+void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > int_data, const NNC& nnc) const {
     const auto& units = this->es.getUnits();
     const IOConfig& ioConfig = this->es.cfg().io();
 
@@ -339,9 +339,9 @@ void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, std::map<st
 
     //Write Integer Vector Map
     {
-        std::map<std::string, std::vector<int> >::iterator it = map.begin();
+        std::map<std::string, std::vector<int> >::iterator it = int_data.begin();
 
-        while(it != map.end())  {
+        while(it != int_data.end())  {
             std::string key = it->first;
             if (key.size() > ECL_STRING8_LENGTH)
               throw std::invalid_argument("Keyword is too long.");
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -220,7 +220,6 @@ EclipseIO::Impl::Impl( const EclipseState& eclipseState,
 {}
 
 
-
 void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > int_data, const NNC& nnc) const {
     const auto& units = this->es.getUnits();
     const IOConfig& ioConfig = this->es.cfg().io();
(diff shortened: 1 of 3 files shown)


opm/output/eclipse/EclipseIO.cpp
```diff
@@ -192,7 +192,7 @@ inline std::string uppercase( std::string x ) {
 class EclipseIO::Impl {
     public:
     Impl( const EclipseState&, EclipseGrid, const Schedule&, const SummaryConfig& );
-        void writeINITFile( const data::Solution& simProps, const NNC& nnc) const;
+        void writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > map, const NNC& nnc) const;
```
joakim-hove: `map` is not a very good name - please call it e.g. `int_data`

opm/output/eclipse/EclipseIO.cpp
```diff
@@ -221,7 +221,7 @@ EclipseIO::Impl::Impl( const EclipseState& eclipseState,
 
 
 
-void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, const NNC& nnc) const {
+void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > map, const NNC& nnc) const {
```
joakim-hove: The `Impl`function can just take a `const std::map<std::string, std::vector<int>>&` argument; observe that the main external function should be called with value semantics (as it is currently) - to allow calling scopy to dispatch this in a separate execution context.

opm/output/eclipse/EclipseIO.cpp
```diff
@@ -337,6 +337,25 @@ void EclipseIO::Impl::writeINITFile( const data::Solution& simProps, const NNC&
     }
 
 
+    //Write Integer Vector Map
+    {
+        std::map<std::string, std::vector<int> >::iterator it = map.begin();
```
joakim-hove: Use range based for loops instead of naked iterators:
```C++
for( const auto& pair : map)  {
     const std::string& key = pair.first;
     const std::vector<int>& int_data = pair.second;
     ...
     if (int_data.size() == this->grid.getCartesianSize( ))
        writeKeyword( fortio , key , int_data );
     else  {
         std::vector<int> global_copy = grid.scatterVector( int_data );
         writeKeyword( fortio , key , global_copy );
     }
}
```

tests/test_EclipseIO.cpp
```diff
@@ -314,9 +317,19 @@ BOOST_AUTO_TEST_CASE(EclipseIOIntegration) {
             { "TRANZ", { measure::transmissibility, tranz, TargetType::INIT } },
         };
 
+        std::map<std::string, std::vector<int> > map;
```
joakim-hove: This can alternatively be initialized as:
```C++
std::map<std::string, std::vector<int>> int_data =  {{"STR_ULONGNAME" , {1,1,1,1,1,1,1,1,1,1,1,1,......}}};
```

tests/test_EclipseIO.cpp
```diff
@@ -343,6 +356,13 @@ BOOST_AUTO_TEST_CASE(EclipseIOIntegration) {
         checkEgridFile( eclGrid );
         loadWells( "FOO.EGRID", "FOO.UNRST" );
 
+        ecl_file_type * ecl_file = ecl_file_open("./FOO.INIT", 0);
```
joakim-hove: No need to add the leading `./`

opm/output/eclipse/EclipseIO.hpp
```diff
@@ -107,7 +107,7 @@ class EclipseIO {
   *     are not yet written to disk.
   */
 
-    void writeInitial( data::Solution simProps = data::Solution(), const NNC& nnc = NNC());
+    void writeInitial( data::Solution simProps = data::Solution(), std::map<std::string, std::vector<int> > map = {}, const NNC& nnc = NNC());
```
joakim-hove: Please add a brief documentation of the new map for integer data, in particular the requirements on the length of the vector:

1. The on-disk keywords will always have `nx*ny*nz` elements.
2. Vectors of length `nactive` will be expanded to length `nx*ny*nz` - inactive cells witll be filled with -1.
3. Others lengths -> exception.

 should be documented - the current choice is not obvious, and people might expect/wish different behavior.

tests/test_EclipseIO.cpp
```diff
@@ -314,9 +317,19 @@ BOOST_AUTO_TEST_CASE(EclipseIOIntegration) {
...
+        std::vector<int> u(27); u[2] = 67; u[5] = 89;
+        map["STR_ULONGNAME"] = u;
+
+        std::vector<int> v(27); v[2] = 67; v[26] = 89;
+        map["STR_V"] = v;
 
         eclWriter.writeInitial( );
-        eclWriter.writeInitial( eGridProps );
+
+        test_assert_throw( eclWriter.writeInitial( eGridProps , map) , std::invalid_argument);
```
joakim-hove: Use the `BOOST_CHECK_xxxx` routines intead of `test_util_xxx`

joakim-hove:
jenkins build this with downstreams opm-parser=1154 please

blattms:
Please try to compile this with OPM/opm-simulators#1336

joakim-hove:
I am deeply sorry - I have misguided regarding vector sizes in this PR. Instead of forcing the output vectors to have `nx*ny*nz`elements the `writeInit( )` method should just pass the integer data directly through to the INIT file - whatever size.

Sorry about the mess.

joakim-hove:
jenkins build this with downstreams opm-simulators=1336 please

blattms:
So for slow thinkers like me: How long does the integer vector has to be?

joakim-hove:
Now - fully unregulated.

blattms:
meaning either number of active cells or number of all cartesian cells
(dims[0]*dims[1]*dims[2]).

opm/output/eclipse/EclipseIO.cpp
```diff
@@ -192,7 +192,7 @@ inline std::string uppercase( std::string x ) {
 class EclipseIO::Impl {
     public:
     Impl( const EclipseState&, EclipseGrid, const Schedule&, const SummaryConfig& );
-        void writeINITFile( const data::Solution& simProps, const NNC& nnc) const;
+        void writeINITFile( const data::Solution& simProps, std::map<std::string, std::vector<int> > int_data, const NNC& nnc) const;
```
atgeirr, line 195: The new argument is taken by value instead of const reference. Is that intentional?

joakim-hove: No - the `Impl` function shoul have been a `const &` - the public function is by intention with value semantics - to support async. That is as with all the IO functions. I suggest we just merge now - this can be fixed with any downstream consequences later.

PR #214, opened 2017-11-15
Well pri 3 keywords

sveinung-r:

PR #217, opened 2017-11-20
Add group oilvapor keywords

lars-petter-hauge:
Added group production rate and totals of free and vaporized oil keywords. ~~I'll rebase after https://github.com/OPM/opm-output/pull/214 is merged.~~ Rebased.

"Free oil production" should be the remaining after subtracting what is vaporized, but the term "Free oil production" isn't really that normal to say, and the only place this is described in the eclipse manual is in fact when "GOPRF" and "GOPTF" is explained.. I'll do some more research before removing the WIP.


lars-petter-hauge:
wip removed, all ready.

PR #218, opened 2017-11-23
Fix FPR and RPR for 2p GasOil case

totto82:


opm/output/eclipse/Summary.cpp
```diff
@@ -375,11 +375,16 @@ quantity fpr( const fn_args& args ) {
 
     const auto& p = args.state.data( "PRESSURE" );
     const auto& pv = args.pv;
-    const auto& sw = args.state.data( "SWAT" );
+    const bool hasSwat = args.state.has( "SWAT" );
+    const auto& swat = hasSwat? args.state.data( "SWAT" ): p;
```
joakim-hove: No; you have to rewrite this (eg using a pointer?); initializing the variable `swat` to pressure if the model does not have `SWAT`is just to error prone.

totto82: Is this better?

joakim-hove:
Personally I would do something like: 

```C++
if (args.state.has("SWAT")) {
    const auto& swat = args.state.data("SWAT");
    for (size_t i = 0; i < p.size(); i++)
          hcs += pv[i] * (1 - swat[i])
} else {
    for (size_t i; i < p.size(); i++)
         hcs += pv[i];
}
```
but if you prefer your variety I will merge that.

What about the test failure downstream?

totto82:
> What about the test failure downstream?

Strange SPE5 doesn't even output FPR.

totto82:
SPE5 fails for everybody. 
I guess we should blame: OPM/ewoms#219

GitPaean:
What happened with Jenkins when testing with that PR?

atgeirr:

PR #220, opened 2017-11-29
wells: add support for energy

andlaus:
This is a rather simple patch which allows to specify energy in the wells. It is completely analogous to solvent and polymer.


joakim-hove:
Is energy flow in wells a real concept?

andlaus:
definitely: if wells don't produce and inject the energy associated with the mass, the reservoir temperature will change *a lot* due to unphysical reasons.

Also note that "heaters" can apparently be simulated by ECL...

PR #221, opened 2017-11-30
WIP:Warn unused sumkeys

lars-petter-hauge:
This is a co-work with sveinung. We were wondering if it would be appropriate with a warning from opm-output if a summary keyword is not handled.

NB: uncertain if simulators test will fail (eclfilecompare, given that the output files may change with this PR). I'm having some issues on my desktop atm.

### DIFF
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -867,7 +867,7 @@ Summary::Summary( const EclipseState& st,
         if (init_config.getRestartRootName().size() <= ECL_STRING8_LENGTH * SUMMARY_RESTART_SIZE)
             restart_case = init_config.getRestartRootName().c_str();
         else
-            OpmLog::warning("Resart case too long - not embedded in SMSPEC file");
+            OpmLog::warning("Restart case too long - not embedded in SMSPEC file");
     }
     ecl_sum.reset( ecl_sum_alloc_restart_writer(basename,
                                                 restart_case,
--- a/opm/output/eclipse/Summary.cpp
+++ b/opm/output/eclipse/Summary.cpp
@@ -883,6 +883,8 @@ Summary::Summary( const EclipseState& st,
     /* register all keywords handlers and pair with the newly-registered ert
      * entry.
      */
+    std::set< std::string > unsupported_keywords;
+
     for( const auto& node : sum ) {
         const auto* keyword = node.keyword();
 
@@ -905,7 +907,10 @@ Summary::Summary( const EclipseState& st,
 
 	    this->handlers->misc_nodes.emplace( keyword, nodeptr ); 
         } else {
-	    if( funs.find( keyword ) == funs.end() ) continue;
+	        if( funs.find( keyword ) == funs.end() ) {
+                unsupported_keywords.insert(keyword);
+                continue;
+            }
 
             if ((node.type() == ECL_SMSPEC_COMPLETION_VAR) || (node.type() == ECL_SMSPEC_BLOCK_VAR)) {
                 int global_index = node.num() - 1;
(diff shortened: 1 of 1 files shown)


opm/output/eclipse/Summary.cpp
```diff
@@ -905,7 +907,10 @@ Summary::Summary( const EclipseState& st,
 
 	    this->handlers->misc_nodes.emplace( keyword, nodeptr ); 
         } else {
-	    if( funs.find( keyword ) == funs.end() ) continue;
+	        if( funs.find( keyword ) == funs.end() ) {
```
joakim-hove, line 910: I guess the call to `OpmLog::info()`could come directly here?

opm/output/eclipse/Summary.cpp
```diff
@@ -940,6 +945,9 @@ Summary::Summary( const EclipseState& st,
 	    this->handlers->handlers.emplace_back( nodeptr, handle );
 	}
     }
+    for ( auto keyword : unsupported_keywords ) {
```
joakim-hove: As suggested I think the log call could come directly; if you prefer to assemble the set the iteration should be:
```C++
for (const auto& keyword: ...
```

alfbr:
> We were wondering if it would be appropriate with a warning from opm-output if a summary keyword is not handled

Absolutely, and this was implemented long ago. Please keep it up-to-date.

opm/output/eclipse/Summary.cpp
lars-petter-hauge: The keyword list includes well-specific items. That is, include "WOPR" and with 3 wells (and WOPR unsupported) you would get 3 warnings (stating that WOPR is unused..).

joakim-hove: Good point

joakim-hove:
> Absolutely, and this was implemented long ago. Please keep it up-to-date.

Ehhh - are you sure about that? The point here is about keywords handled by opm-parser, and then subsequently not handled by opm-output. I was not aware of warnings in this case, but I might just have mislaid my brain.

opm/output/eclipse/Summary.cpp
lars-petter-hauge: Of course, I'll fix this. I'm trying to improve and get rid of those annoying obvious mistakes..

alfbr:
> Ehhh - are you sure about that?

I will need to check, it is a after all a long time ago. But that was certainly the intent at the time. If you run Norne straight from trunk, you will get output like this:

Warning: Keyword 'BASIC' is unhandled for output to file.

Warning: Keyword 'CPU' is unhandled for output to file.

Warning: Keyword 'FIP' is unhandled for output to file.

lars-petter-hauge:
Warning of unhandled restart keywords (e.g. BASIC) have been implemented in opm-simulators. Summary keywords however will not provide a warning. I'm not sure if every other section contain warning about unsupported keywords.

alfbr:
True, you do have the:

Warning: Keyword 'NEWTON' is unhandled for output to file.

which I thought was from the summary section. However, I will need to look back to see what was actually done.

lars-petter-hauge:
Fixed comment from Joakim, so this should be ready now

lars-petter-hauge:
> Warning: Keyword 'NEWTON' is unhandled for output to file.

I suspect that the reason is NEWTON exists both as output and input. That is, the warning generated is based on NEWTON=X set in SCHEDULE, not if NEWTON is added to SUMMARY section.

joakim-hove:
> NB: uncertain if simulators test will fail (eclfilecompare, given that the output files may change with this PR). I'm having some issues on my desktop atm.

Integration tests will only consider the binary output files, so this PR should not induce any changes.

lars-petter-hauge:
> Integration tests will only consider the binary output files, so this PR should not induce any changes.

Thanks. I saw "compare ECLfiles" in the opm-simulators, so I got a little bit uncertain whether or not it would fail (of course I agree that comparing anything but binary files would be hopeless..)

alfbr:
> I suspect that the reason is NEWTON exists both as output and input

Sounds probable. I will need to go back and look at the actual implementation. Generally speaking though, we do want warnings whenever a user requests a vector in the summary section that is not supported. Since the output code ideally supports being used by multiple simulators, this logging may belong on the simulator side.

PR #226, opened 2017-12-19
changed: put prereqs in module itself

akva2:
Upstream: https://github.com/OPM/opm-common/pull/304


joakim-hove:
This can be merged independently of upstream?

akva2:
travis seems to imply that yes. however without the upstream changes it's basically a NOP.

joakim-hove:
Ok -I'll wait for upstream, and let you merge it when appropriate.

PR #228, opened 2018-01-05
fixed: package prereqs file

akva2:


akva2:
i'm going to self-merge this series. nobody really reviews and this will restore operational nightly builds.

Issue #229, opened 2018-01-15
Support for FPRP


berland:
Support for the summary keyword FPR (hydrocarbon pore weighed average) is implemented, but the simpler FPRP (pore volume weighted average) is missing. 

`fpr()` in `Summary.cpp` implements `sum(HCPV*p)/sum(HCPV)` while FPRP would need only `sum(PV*p)/sum(PV)`.

New function `fprp()` or should it be templatized?

PR #230, opened 2018-01-16
Add support for FPRP

lars-petter-hauge:
ref: https://github.com/OPM/opm-output/issues/229

### DIFF
--- a/tests/test_Summary.cpp
+++ b/tests/test_Summary.cpp
@@ -1063,6 +1063,43 @@ BOOST_AUTO_TEST_CASE(fpr) {
     BOOST_CHECK_CLOSE( fpr2_si, ecl_sum_get_field_var( resp2, 1, "FPR" ), 1e-5 ); //
 }
 
+BOOST_AUTO_TEST_CASE(fprp) {
+    setup cfg( "test_fprp", "summary_deck_non_constant_porosity.DATA");
+
+    out::Summary writer( cfg.es, cfg.config, cfg.grid, cfg.schedule, cfg.name );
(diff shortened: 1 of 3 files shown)


tests/test_Summary.cpp
```diff
@@ -1063,6 +1063,51 @@ BOOST_AUTO_TEST_CASE(fpr) {
...
+
+    UnitSystem units( UnitSystem::UnitType::UNIT_TYPE_METRIC );
+    // fprp = sum_ (p * pv ) / pv
+    const double fprp_si =  ( (3 * 0.1 + 8 * 0.2) * 500 ) / ( (500*0.1 + 500*0.2));
+    const double fprp = units.from_si( UnitSystem::measure::pressure, fprp_si );
+    BOOST_CHECK_CLOSE( fprp, ecl_sum_get_field_var( resp, 1, "FPRP" ), 1e-5 ); //
+
+    // change pressure and test again.
+    size_t g = 0;
+    for (size_t k=0; k < cfg.grid.getNZ(); k++) {
```
joakim-hove: This is of course perfectly OK - but to to be slightly anal I would suggest looping over the active index instead:
```C++
for (size_t g=0; g < grid.getNumActive(); g++)
     cfg.solution.data("PRESSURE")[g] = units.to_si( ... )
```
In the input and output handling there is certainly some use of the cartesian indices, but in the simulator as such there is really no concept of a cartesian coordinate system and all sums are only over active cells.

joakim-hove:
Merge when/if you have updated the test, or ping me here and I will do it.

lars-petter-hauge:
Thanks for the review. Not only is it more readable, it reduces number of lines and variables as well (no need for triple-for loop). Guess I'm a bit too attached to thinking how cell properties are set..

PR #234, opened 2018-01-19
changed: use PROJECT_BINARY_DIR instead of CMAKE_BINARY_DIR

akva2:
to aid use of a super project for opm, see https://github.com/OPM/opm-common/pull/321


joakim-hove:
What is the order merging between this and  OPM/opm-common#321?

akva2:
completely independent, this makes no difference to current build (CMAKE_BINARY_DIR and PROJECT_BINARY_DIR is the same when there's only a single project).

PR #235, opened 2018-01-22
Get block and region values from the simulator 

totto82:
Remove the computations in opm-output that depends on cell values. 

Motivation: 
Avoid duplication of code. 
Speed up MPI run. Now calculations are done locally and only the block, region and field values are communicated. 

Depends on downstream code.

### DIFF
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -409,10 +409,10 @@ void EclipseIO::writeTimeStep(int report_step,
                               double secs_elapsed,
                               data::Solution cells,
                               data::Wells wells,
-                              const std::map<std::__cxx11::string, double>& single_summary_values,
-                              const std::map<std::__cxx11::string, std::vector<double> >& region_summary_values,
-                              const std::map<std::pair<std::__cxx11::string, int>, double>& block_summary_values,
-                              const std::map<std::__cxx11::string, std::vector<double>>&extra_restart,
+                              const std::map<std::string, double>& single_summary_values,
+                              const std::map<std::string, std::vector<double> >& region_summary_values,
+                              const std::map<std::pair<std::string, int>, double>& block_summary_values,
+                              const std::map<std::string, std::vector<double>>& extra_restart,
                               bool write_double)
  {
 
@@ -438,7 +438,6 @@ void EclipseIO::writeTimeStep(int report_step,
                                           es,
                                           schedule,
                                           wells ,
-                                          cells ,
                                           single_summary_values ,
                                           region_summary_values,
                                           block_summary_values);
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -401,13 +401,6 @@ void EclipseIO::writeInitial( data::Solution simProps, std::map<std::string, std
             this->impl->writeEGRIDFile( nnc );
     }
 
-    this->impl->summary.set_initial( simProps );
-}
-
-
-void  EclipseIO::overwriteInitialOIP( const data::Solution& simProps )
-{
-    this->impl->summary.set_initial( simProps );
 }
 
 // implementation of the writeTimeStep method
@@ -419,7 +412,7 @@ void EclipseIO::writeTimeStep(int report_step,
                               std::map<std::string, double> misc_summary_values,
                               std::map<std::string, std::vector<double>> region_summary_values,
                               std::map<std::string, std::vector<double>> extra_restart,
-			      bool write_double)
+                              bool write_double)
  {
 
     if( !this->impl->output_enabled )
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -409,10 +409,10 @@ void EclipseIO::writeTimeStep(int report_step,
                               double secs_elapsed,
                               data::Solution cells,
                               data::Wells wells,
-                              std::map<std::string, double> single_summary_values,
-                              std::map<std::string, std::vector<double>> region_summary_values,
-                              std::map<std::pair<std::string, int>, double> block_summary_values,
-                              std::map<std::string, std::vector<double>> extra_restart,
+                              const std::map<std::__cxx11::string, double>& single_summary_values,
+                              const std::map<std::__cxx11::string, std::vector<double> >& region_summary_values,
+                              const std::map<std::pair<std::__cxx11::string, int>, double>& block_summary_values,
+                              const std::map<std::__cxx11::string, std::vector<double>>&extra_restart,
                               bool write_double)
  {
 
--- a/opm/output/eclipse/EclipseIO.cpp
+++ b/opm/output/eclipse/EclipseIO.cpp
@@ -417,6 +417,7 @@ void EclipseIO::writeTimeStep(int report_step,
                               data::Solution cells,
                               data::Wells wells,
                               std::map<std::string, double> misc_summary_values,
+                              std::map<std::string, std::vector<double>> region_summary_values,
                               std::map<std::string, std::vector<double>> extra_restart,
 			      bool write_double)
  {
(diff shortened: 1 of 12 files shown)


joakim-hove:

joakim-hove:
You have been really deep inside the code - impressive; I hope your mental health has not been to severely affected.

joakim-hove:
OK - this is a large change; I guess there will be some iterations in the combined task before it is merged. I see the point from a performance point of view - but is it worth it? This way quite a lot of Eclipse specific output details leek all the way out `ewoms` - your call!

Sepcifically the PR implies a certain change in responsability between the simulator and the output layer; consider e.g. the summary keyword `RPR`. Previously the loop was fully the reponsability of the output layer, now this is the responsability of the simulator to correctly initialize the `region_vars` for all region values, and the  - output layer can not detect missing regions. Not necessarily a problem?

joakim-hove:
As part of the API change the tests here in OPM output have been updated - very good; however quite a lot of functionality has now been moved to ewoms and I would strongly encourage you to reimplement the tests there.

joakim-hove:
I am not certain about this, but my gut feeling is that you could simplify (i.e. remove code) even further:

1. Maybe the `data::Solution cells`argument could be completely removed?
2. I guess the  `RegionCache` class -can be simplified?

opm/output/eclipse/Summary.cpp
```diff
@@ -1111,30 +941,49 @@ void Summary::add_timestep( int report_step,
...
-	if (node_pair != this->handlers->misc_nodes.end()) {
-	    const auto * nodeptr = node_pair->second;
-	    const auto unit = misc_units.at( key );
-	    double si_value = value_pair.second;
-	    double output_value = es.getUnits().from_si(unit , si_value );
-	    ecl_sum_tstep_set_from_node( tstep, nodeptr , output_value );
-	}
+        for (size_t reg = 0; reg < value_pair.second.size(); ++reg) {
+            const auto node_pair = this->handlers->region_nodes.find( std::make_pair(key, reg+1) );
+            if (node_pair != this->handlers->region_nodes.end()) {
```
joakim-hove, line 976: I guess if this `if` test does not apply we should at least have a warning?

totto82:
> I see the point from a performance point of view - but is it worth it?

Most of the code for the regions part was already in place since it is needed for the .PRT output. I just needed to add a way to communicated it to opm-output. For the block part I needed extra code, but the upside on the parallel performance is even larger. In the old code, lot of cell values where gathered from the other nodes at every substep. Now only the actually values are communicated. I wasn't (on till now) more than I days work, so I think it was worth it. 


> As part of the API change the tests here in OPM output have been updated - very good; however quite a lot of functionality has now been moved to ewoms and I would strongly encourage you to reimplement the tests there.

Currently we depend on the integration tests for this. 


>I am not certain about this, but my gut feeling is that you could simplify (i.e. remove code) even further:
>   Maybe the data::Solution cellsargument could be completely removed?
>   I guess the RegionCache class -can be simplified?

I don't have the overview of this code so I didn't clean more than I what was obvious.

totto82:
In `RestartConfig::handleScheduleSection(...)` in [`RestartConfig.cpp`](https://github.com/OPM/opm-parser/blob/master/lib/eclipse/EclipseState/IOConfig/RestartConfig.cpp#L379) it seems like RPTRST and RPTSCHED are handled the same way. i.e. added to the same keyword list. Shouldn't separate list be kept for these keywords since they control different things? When I do `getRestartKeyword(..)` I also get RPTSCHED keywords.

joakim-hove:
> so I think it was worth it.

Good - I was actually thinking more of the spreading of the details reated to Eclipse output into the Ewoms code - but I guess the conclusion is the same.

joakim-hove:
>> [...]  been moved to ewoms and I would strongly encourage you to reimplement the tests there.

>Currently we depend on the integration tests for this.

OK - I think it is unfortunate that the test coverage of flow is reduced and would encourage you to reconsider - I would be happy to assist you in the porting. This specially applies since this is now implemented with MPI - something which is notoriously difficult.

joakim-hove:
> I don't have the overview of this code so I didn't clean more than I what was obvious.

I understand that - but please look into the possibility of removing the `data::Solution cells` variable; would be good to remove it - *iff* it is trivial.

joakim-hove:
> [...] Shouldn't separate list be kept for these keywords since they control different things? 

There might be bugs of course; but configuring output is notoriously ugly - and spesifcially both the mentioned keywords configure what should go in he restart files. In different ways - see [here](https://github.com/OPM/opm-parser/blob/master/lib/eclipse/include/opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp#L28)

totto82:
> There might be bugs of course; but configuring output is notoriously ugly - and spesifcially both the mentioned keywords configure what should go in he restart files. In different ways - see here

This is ugly......

What I want is to avoid creating the cell values. Adding something like this
```C++
if (!restartConfig.getWriteRestartFile(reportStepNum) || substep)
   return; 
```
Currently I have to pass RESTART_SOLUTION values for all but the substeps.

It seems like the RPTSCHED keyword mess the logic up.  I can try to investigate this a bit to see whether this is a bug in my head or in the code.

opm/output/eclipse/EclipseIO.cpp
```diff
@@ -416,9 +409,11 @@ void EclipseIO::writeTimeStep(int report_step,
                               double secs_elapsed,
                               data::Solution cells,
                               data::Wells wells,
-                              std::map<std::string, double> misc_summary_values,
-                              std::map<std::string, std::vector<double>> extra_restart,
-			      bool write_double)
+                              const std::map<std::__cxx11::string, double>& single_summary_values,
```
bska: We can't use symbols like `std::__cxx11::string` in our interfaces.  Those are in the implementation's (i.e., compiler + library's) namespace.  Is there a reason `std::string` cannot be used?

totto82: I didn't notice. I must have been changed automatically by qt-creator. It has a convenient (but it seems not very bullet prof) option for synchronizing the .cpp and .hpp file. I will revert this back to std::string.

joakim-hove:
>What I want is to avoid creating the cell values. Adding something like this

```C++
if (!restartConfig.getWriteRestartFile(reportStepNum) || substep)
   return; 
```
I don't really understand what you mean - maybe a &#9742; is required?

joakim-hove:
> What I want is to avoid creating the cell values. 

Maybe this [`RestartConfig.getRestartKeywords( )`](https://github.com/OPM/opm-parser/blob/master/lib/eclipse/include/opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp#L328) is the method you are after? I do not think that has been used in flow yet?

totto82:
> Maybe this RestartConfig.getRestartKeywords( ) is the method you are after? I do not think that has been used in flow yet?

I use this in the new code for data of type: Opm::data::TargetType::RESTART_AUXILIARY. 

> I don't really understand what you mean - maybe a ☎ is required?

I found the issue. "PRESSURE", "SWAT" etc is used for RFT output. The current code adds these field  unconditionally for all report steps regardless of getWriteRestartFile(). A proper fix would be to add the connection pressure and saturations to the COMPLETION struct in Wells.hpp in opm-output.

joakim-hove:
> A proper fix would be to add the connection pressure and saturations to the COMPLETION struct in Wells.hpp in opm-output.

Makes sense - or you on to this?

joakim-hove:
I am happy to merge the opm-output part of this; although I do find it slightly surprising you are so motivated drag Eclipse peculiarities out of the opm-output realms.

andlaus:
We should probably not be too dogmatic about such stuff and go with the simplest solution whenever reasonable: what is `flow` if not a big collection of Eclipse peculiarities? (maybe we should consider renaming it to "eclweirdnessreproducer" or something ;))

opm/output/eclipse/Summary.cpp
```diff
@@ -956,7 +749,11 @@ class Summary::keyword_handlers {
     public:
         using fn = ofun;
         std::vector< std::pair< smspec_node_type*, fn > > handlers;
-        std::map< std::string, smspec_node_type* > misc_nodes;
+        std::map< std::string, smspec_node_type* > singel_value_nodes;
```
atgeirr: Typo: singel -> single

PR #238, opened 2018-01-25
Res vol keywords (WVPR + WVPT & WVPRT+FVPRT)

lars-petter-hauge:

PR #242, opened 2018-02-01
Added braces to for loop.

joakim-hove:
The purpose of this PR is just to serve as a handle to initiate an update of the test data. 

Sorry about the f... ups.

### DIFF
--- a/tests/test_Restart.cpp
+++ b/tests/test_Restart.cpp
@@ -385,8 +385,9 @@ void compare( const RestartValue& fst,
         if (pair.first == "TEMP")
             tol *= 10;
 
-        for( ; first != fst.solution.data( pair.first ).end(); ++first, ++second )
+        for( ; first != fst.solution.data( pair.first ).end(); ++first, ++second ) {
             BOOST_CHECK_CLOSE( *first, *second, tol );
+        }
     }
 
     BOOST_CHECK_EQUAL( fst.wells, snd.wells );


joakim-hove:
jenkins build this with downstreams update_data please

akva2:
jenkins build this update_data please

atgeirr:
I am not convinced that it is sufficient to only update data. I will discuss this in OPM/opm-parser#1203

joakim-hove:
jenkins build this update_data please

joakim-hove:
jenkins build this opm-data=299 please

GitPaean:
> Existing PR https://github.com/OPM/opm-data/pull/ 299 was updated


It looks like the extra space in the link makes the link will not work. @akva2

joakim-hove:
OK - yesterday I went on a merging spree and really really fucked up. I merged three different pull requests:

1. https://github.com/OPM/opm-parser/pull/1203
2. https://github.com/OPM/opm-output/pull/238
3. https://github.com/OPM/opm-output/pull/239

All of these should have been accompanied with updates to the test data, none of theme were. The PR in opm-parser has subsequently been reverted in https://github.com/OPM/opm-parser/pull/1204 - as that might have had real issues - not yet clear. The PR's in opm-output were quite trivial - and "obviosuly correct" -  but they did add new keywords to the summary files and should have been accompanied with a data update.

With this place-holder PR I have now updated the data, and created https://github.com/OPM/opm-data/pull/299[1] When jenkins is green I think this should be ready to work, but since my fat fingers have been messing up all over the place lately I would prefer an extra pair of eyes before merging this and https://github.com/OPM/opm-data/pull/299 @akva2 / @atgeirr?

 I would have been more comfortable if the PR only contained changes to the summary files `.SMSPEC` and `.UNSMRY` as those changes can be directly ascribed to the two PR's in opm-output, whereas I do not understand where the changes in the `.UNRST`files come from? Can that be a left-over/pollution from the first `update_data` build on this PR? Have started a clean place-holder PR in #243

[1]: Regarding the [comment](https://github.com/OPM/opm-data/pull/299#issuecomment-362260274) - this PR should now only come from the changes due to the two PR's in opm-output, but when it was initially created - and @atgeirr commented - it also contained the `TEMP` changes.

PR #250, opened 2018-02-08
Some more cleaning in the output code  in opm-output, ewoms and opm-simulator 

totto82:
1) Add well communication to data::wells. 
2) Pass cell data used by RFT via the data::wells::completion. 
Motivation: Avoid communication of all cell data. 
3) Bugfix. Store global index instead of active index in completions. RFT data is wrong for MPI simulations on current master. 

Merge together with downstream PRs.

### DIFF
--- a/opm/output/data/Wells.hpp
+++ b/opm/output/data/Wells.hpp
@@ -102,10 +102,10 @@ namespace Opm {
     };
 
     struct Completion {
-        using active_index = size_t;
+        using global_index = size_t;
         static const constexpr int restart_size = 2;
 
-        active_index index;
+        global_index index;
         Rates rates;
         double pressure;
         double reservoir_rate;
@@ -146,7 +146,7 @@ namespace Opm {
         }
 
 
-        double get(const std::string& well_name , Completion::active_index completion_grid_index, Rates::opt m) const {
+        double get(const std::string& well_name , Completion::global_index completion_grid_index, Rates::opt m) const {
             const auto& witr = this->find( well_name );
             if( witr == this->end() ) return 0.0;
 
--- a/opm/output/data/Wells.hpp
+++ b/opm/output/data/Wells.hpp
@@ -109,6 +109,9 @@ namespace Opm {
         Rates rates;
         double pressure;
         double reservoir_rate;
+        double cell_pressure;
+        double cell_saturation_water;
+        double cell_saturation_gas;
 
         template <class MessageBufferType>
         void write(MessageBufferType& buffer) const;
@@ -290,6 +293,9 @@ namespace Opm {
             this->rates.write(buffer);
             buffer.write(this->pressure);
             buffer.write(this->reservoir_rate);
+            buffer.write(this->cell_pressure);
+            buffer.write(this->cell_saturation_water);
+            buffer.write(this->cell_saturation_gas);
     }
 
     template <class MessageBufferType>
@@ -327,6 +333,9 @@ namespace Opm {
             this->rates.read(buffer);
             buffer.read(this->pressure);
             buffer.read(this->reservoir_rate);
+            buffer.read(this->cell_pressure);
+            buffer.read(this->cell_saturation_water);
+            buffer.read(this->cell_saturation_gas);
    }
 
     template <class MessageBufferType>
--- a/opm/output/data/Wells.hpp
+++ b/opm/output/data/Wells.hpp
@@ -77,6 +77,11 @@ namespace Opm {
             /// true if any option is set; false otherwise
             inline bool any() const noexcept;
 
+            template <class MessageBufferType>
+            void write(MessageBufferType& buffer) const;
+            template <class MessageBufferType>
+            void read(MessageBufferType& buffer);
+
         private:
             double& get_ref( opt );
             const double& get_ref( opt ) const;
@@ -104,6 +109,11 @@ namespace Opm {
         Rates rates;
         double pressure;
         double reservoir_rate;
+
+        template <class MessageBufferType>
+        void write(MessageBufferType& buffer) const;
+        template <class MessageBufferType>
+        void read(MessageBufferType& buffer);
     };
 
     struct Well {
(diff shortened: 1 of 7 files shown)


opm/output/data/Wells.hpp
```diff
@@ -145,9 +162,35 @@ namespace Opm {
             return completion->rates.get( m, 0.0 );
         }
 
+        template <class MessageBufferType>
```
joakim-hove, line 165: Where did this `MessageBufferType` come from?

joakim-hove:
To me this looks like a quite large change - more than a post merge cleanup? At least the transfer through the `MessageTypeBuffer` comes a bit out of the blue?!

opm/output/data/Wells.hpp
totto82: It is set in collecttoiorank.hh in ewoms/ebos
```C++
typedef Dune :: Point2PointCommunicator< Dune :: SimpleMessageBuffer > P2PCommunicatorType;
typedef typename P2PCommunicatorType :: MessageBufferType MessageBufferType;
```
I wanted to add the communication for the `data::well` in the same class to make it easier to remember to add the communication part when adding new data to `data::wells`. I have added a test in ebos (`test_ecl_output.cc`) to make sure everything is communicated.

totto82:
> To me this looks like a quite large change - more than a post merge cleanup? 

It is supposed to be more than a post merge cleanup. It is more like the final step in my work of re-distributing the work for output between opm-output, ewoms and opm-simulator. With this the output code almost scales (only tested up to 8 cores on my machine). Before it used to increase with increased number of nodes.

joakim-hove:
>[...]  It is more like the final step in my work of re-distributing the work for output between opm-output, ewoms and opm-simulator. 

OK

>With this the output code almost scales (only tested up to 8 cores on my machine). Before it used to increase with increased number of nodes.

Well - that is certainly impressive!

joakim-hove:
jenkins build this please

opm/output/data/Wells.hpp
joakim-hove: I **really really** dislike this way of tricking the dependency graph using templates. ewoms depends on opm-ouptut. now you introduce symbols in opm-output which makes opm-output depend on ewoms in a circular fashion, and the only reason this works is that the code in opm-output is dead and not accessed before we reach ewoms. I will not merge this in the present form.

My main gripe with the current code is that you introduced a hidden dependency - and now have methods in the well object *which can not be tested*; I certainly see the rationale for doing it like this - but then we should in my opinion just suck it up - and make opm-output depend explicitly on the dune communicator type - because that is now the case. A minimum solution would be to move the `test_ecl_output.cc` from ebos to opm-output and then either make the test implementation depend explicitly on the dune commonicator, or use a mock messagebuffer - I guess `std::stringstream` would get you far.

atgeirr: To make opm-output depend on the Dune communicator should be avoided. I realize now that this changes the thinking that opm-output is parallel-unaware, and we should not do that lightly. The minimum solution suggested by @joakim-hove is ok (with a trivial messagebuffer used for testing, not depending on the ewoms class or moving it), but I think there may be better solutions. The simplest alternative is to put the `read()` and `write()` methods as free functions in ebos, although that would not resolve the problem you try to solve here (that the class gets more members and the buffer read/write function updates are forgotten).

totto82: Ok, I see your points. With the current solution in `data::well` I really don't like to move the `read()` and `write()` methods to ebos. A more generic `data::well` container, more like `data::solution`, would make moving the `read()` and `write()` methods to ebos less problematic in my opinion, then the communication code could be generic as well without need for frequent updates. This would certainly require a lot of work both in the opm-output code and in ebos. If the minimum solution is acceptable for both of you, I would like to go for that. @joakim-hove I guess you don't want me to move the whole `test_ecl_output.cc` to opm-output, but the test that explicitly tests the `read()` and `write()` methods. For that I think extending the test_summary is sufficient.

atgeirr: I have an idea that could work, but I have not thought it 100% through:

To separate "accessing specific data members" from parallel communication, the opm-output classes could define `operator<<` and `operator>>`, (alternatively call them `read()` and `write()`) which would be completely normal I/O functions, writing to standard streams. I would suggest ASCII is acceptable for the limited amount of data related to wells, but we could make it binary. Then the communication code just uses `stringstreams` to get in/out these data (that part does not care about the contents other than it being X bytes).

To build the global object, we would need the `read()` method to append, not overwrite the existing content. So perhaps we should call it `append()` instead.

Bonus benefit: easier print debugging. (Not a big one...)

dr-robertk: ASCII is essentially not acceptable here. And then your suggestions with `stringstream` becomes quite cumbersome and error prone. @totto82 suggestion is quite simple and easy to understand. The template argument with `MessageBuffer` can be easily tested locally.

atgeirr: > The template argument with MessageBuffer can be easily tested locally.

With a trivial MessageBuffer class I assume? That is an acceptable solution for me as I wrote earlier, I just tried to see if there were alternative solutions.

totto82: I have added read/write test in test_summary. Is this what you had in mind?

joakim-hove: > I have added read/write test in `test_summary`. Is this what you had in mind?

Yes - this was what I had in mind.

joakim-hove:
OK; with test added to `test_summary` I am happy with this.

opm/output/eclipse/Summary.cpp
```diff
@@ -727,6 +727,15 @@ static const std::unordered_map< std::string, UnitSystem::measure> region_units
   {"RWIP"     , UnitSystem::measure::volume }
 };
 
+static const std::unordered_map< std::string, UnitSystem::measure> block_units = {
```
akva2: remove or keep? seems the conflict resolution here took the wrong blocks?

PR #251, opened 2018-02-09
merging three reservoir phase rates to be one reservoir volume rate.

GitPaean:
We do not have keywords to support the reservoir phase rates. 

Voidage rate can be another name for it. Not sure which one is better.  Please suggest if you have an opinion.


GitPaean:
It turned out there will be reservoir phase rates. They are `WOVPR`, `WOVIR`, and `WWVPR`, `WWVIR`,  and `WGVPR`, `WGVIR`.  Closing this PR now.

GitPaean:
A correction, only WWVIR, WGVPR and WGVIR are related, other keywords are not related though.

PR #252, opened 2018-02-12
fixed: libecl recently enforced stricter filename rules

akva2:
in particular, no mixed case. either all upper or all lower.

i'm not sure what the reason is, but here's the monkey work.


joakim-hove:
jenkins build this please

joakim-hove:
Thank you for the fix - and the heads up. This was on my plate - but something happened before I could complete my approach.

The thing is that Eclipse assumes (actually fails hard if not) that the datafile - and all binary output files -  have EITHER_ALL_UPPERCASE or all_lowercase - MixedCase is not allowed. I guess the reason is  to offer guaranteed behaviour on case insensitive filesystems.

The low level functions for IO in libecl follow normal platform semantics on this, but after: https://github.com/Statoil/libecl/pull/308 the functions which *generate* an Eclipse filename will fail if the input basename has mixed case. That applied to many of the tests in opm-output, which then failed.

If you have updated libecl you need to update opm-output with the fix applied here, but it is not the other way around - you do *not* need to update libecl.

atgeirr:
Thanks for the explanation! I guess the functions in questions return an error code that we should check eventually?

lindkvis:
Hi,

Should the ecl_util_select_filelist not return some sort of error if the base name was not valid?

Although I suppose it is possible to call ecl_util_valid_basename() first.

andlaus:
note that the `opm-output` module is deprecated. If you expect to get answers, I guess you should point any questions to either https://github.com/OPM/opm-common or to https://github.com/Statoil/libecl because people may not get notifications for stuff posted here anymore.

PR #257, opened 2018-03-05
WIP: Match timestep for well info

lars-petter-hauge:
This PR handles when opm-output takes into account changes in group policy or efficiency factor during a run. Opm-output would previously change those configurations according to when they are parsed, however as a reporting tool, the configuration should instead be set one timestep later.

I am not too fond of how this PR looks and where timesteps are changed. Suggestions for other implementations are welcome.

Increment timestep where necessary to avoid premature well configuration.

Consider the following:
```
.
.
.

WELLSPECS
 'W1'  'G1' ... /

TSTEP
 2*1 /

WELLSPECS
 'W1'  'G2' .../

TSTEP
 2*1 /
```
In that case the parent group of W1 is changed after two timesteps. At timestep two the Group parent will indicate G2, while the GOPT:G2 at timestep two should be reported for what has progressed up until timestep two.


lars-petter-hauge:
Note that this pr requires update for SPE9_CP_GROUP test in opm-simulators (it contains GEFAC that is added later than the first time step).

joakim-hove:
I will try to look at this later today - but my initial gut feeling was that this should have been in the confiuration layer in opm-parser instead.

atgeirr:
New PR must be opened in opm-common if this feature is to be included.

lars-petter-hauge:
Yepp, I'll add it to opm-common instead

